PR #274, opened 2021-07-13
Handling client connection

fisyamaloy:
1) Connection data to server was announced 2 times, then i moved it in another place and take data from ServerInfo.hpp.
2) Invalid connections are handled properly and users are notified.

### DIFF
--- a/Client/Client.Core/Client.Core.Static/ServerInfo.hpp
+++ b/Client/Client.Core/Client.Core.Static/ServerInfo.hpp
@@ -2,6 +2,6 @@
 
 namespace ServerInfo
 {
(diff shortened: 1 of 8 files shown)


Client/Client.Core/Client.Core.Static/ServerInfo.hpp
```diff
@@ -0,0 +1,7 @@
+#include <string>
+
+namespace ServerInfo
+{
+	const std::string address = "104.40.239.183";
```
f1nal3, line 5: Both can be constexpr, first one has to be string view though

StasDash (approved):

Svidunovich (approved):

PR #296, opened 2021-07-16
Moved SQLite to a different target

f1nal3:
This way server doesn't get the Qt dependency


Orange-hanter (approved):

DataAccess/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
 project(DataAccess LANGUAGES CXX)
 
 add_subdirectory(Public)
-add_subdirectory(DataAccess.Static)
 add_subdirectory(DataAccess.Debug)
+add_subdirectory(DataAccess.Static)
```
vDrWool, line 5: Maybe then rename to DataAccess.Postgre?

f1nal3: yeah, makes sense

f1nal3: done.

Svidunovich (approved):

PR #299, opened 2021-07-20
Basic Authentication

f1nal3:


f1nal3:
At this point on properly entered credentials login occurs. Wrong credentials get a bad response and that is subject of change. Next step is to tie userID with the connection after login is successful.

vDrWool (approved):

StasDash (approved):

PR #310, opened 2021-07-22
Stress tester - WIP

vDrWool:
Performance Tester helps us to check for some 'jump' in performance.
It creates some instances of app that we need to check and pass from 25 to 75 args (like 'cl', 'mh', etc).

Unfortunatly, for now **perfTest is working via 'crutch'**. I added a delay of 5 seconds for the client, otherwise the client crashes when passing arguments to it. **This have to be solved somehow.**

And also I need that _someone check it on Linux and MacOS_.


Svidunovich:
Do we need all the comments in this pr?

vDrWool:
> Do we need all the comments in this pr?
> If not, fix it
> plz

vDrWool:
Comments in Tester.cpp I won't to delete becase they are about how it works (I guess it a bad idea).

PerformanceTest/ArgumentParser.cpp
```diff
@@ -0,0 +1,92 @@
...
+	{
+		switch (arg)
+		{
+		case argument::path:
+			return fs::exists(value);
+		case argument::count:
+			return !value.empty()
+				&& (std::atoi(value.data()) < _MAX_COUNT);
+				// WHY IS IT NOT COMPILING!??!!?
+				//&& std::all_of(value.cbegin(), value.cend(), std::isdigit);
```
Svidunovich, line 27: Do we need these comments?

PerformanceTest/SimpleTimer.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+	void SimpleTimer::stop() noexcept 
+	{ 
+		_end = hsc::now();
+	}
+
+	void SimpleTimer::reset() noexcept 
+	{ 
+		_start        = {};
+        _end          = {};
```
Svidunovich, line 20: Please check spaces in all documents

Svidunovich (changes requested):

Svidunovich (approved):

StasDash (approved):

GuskoMaksim (approved):

Issue #272, opened 2021-07-12
Messages not being attached to channels.


f1nal3:
Currently if you change active channel the messages do not change accordingly.
This is because messages and channels are not tied together.

Implement a way to do this
@f1nal3 will add some test entries into the database.

[AC]: Changing channels in Qt Client shows only messages belonging to that channel.

PR #311, opened 2021-07-23
connect channels with chat

Svidunovich:
Added switching between channels.
#272


Client/Client.Qt/ChatWindow.cpp
```diff
@@ -8,14 +8,26 @@ ChatWindow::ChatWindow(QWidget* parent) : QWidget(parent)
...
-    _mainLayout->setStretchFactor(0, 20);
-    _mainLayout->setStretchFactor(1, 80);
+    _mainLayout->addWidget(_chatSwitchWidget.get());
+
+    connect(_channelListWidget->getChannelList().get(), &ListWidget::itemPressed,
+            [&](){
+            if(_chatSwitchWidget->count() == _channelListWidget->getChannelList()->currentRow())
+            {
+                _chatSwitchWidget->insertWidget(_channelListWidget->getChannelList()->currentRow(), new ChatWidget());
+                _mainLayout->setStretchFactor(1, 80);
```
StasDash, line 23: Do not change streatch factor on every single chat. Set it once:
_mainLayout->setStretchFactor(0, 20);
_mainLayout->setStretchFactor(1, 80);
This is because user can drag splitter to change it manually

StasDash (changes requested):

Client/Client.Qt/ChatWindow.cpp
Svidunovich: Doesn't it change 1 time?
Because if I change 1 chat and add the 2nd, changes will remain

StasDash: You change user's stretch factor on chat change. It mustn't happen.

StasDash (approved):

vDrWool (approved):

Issue #210, opened 2021-05-11
Improve the message processing mechanism on the Client-side. 


glaush:
The purpose of this task is to improve the unpacking of messages from the queue, followed by processing in the different callbacks.

At the moment we are unpacking messages from the queue in this way:

![image](https://user-images.githubusercontent.com/56369021/117843285-5ae1ab00-b287-11eb-9ee3-052750ffd1e1.png)

We need something of that: 

![image](https://user-images.githubusercontent.com/56369021/117844660-89ac5100-b288-11eb-924a-269610c2e1a7.png)

Orange-hanter:
#313 
test

PR #313, opened 2021-07-26
Fixing the sequence of client connection and window creation

GuskoMaksim:


Client/Client.Qt/Widgets/ChannelListWindow.cpp
```diff
@@ -28,7 +28,11 @@ ChannelListWindow::ChannelListWindow(std::shared_ptr<ListWidget>& anotherChannel
 
     _addChannelButton->setClickCallback([this]() { addChannelToMainChannelWidget(); });
     _updateChannelButton->setClickCallback([this]() { updateChannelListWindow(); });
-    ConnectionManager::getClient().askForChannelList();
+    if (ConnectionManager::isConnected())
+    {
+        ConnectionManager::getClient().askForChannelList();
+
```
Orange-hanter, line 34: Please, follow code style rules

Svidunovich (approved):

Orange-hanter (changes requested):

Client/Client.Qt/Application.cpp
```diff
@@ -56,6 +56,7 @@ void Application::setAppState(App::AppState app_state)
         }
         case App::AppState::ChatWindowForm:
         {
+            _mainwidget->addWidget(std::make_unique<ChatWindow>());
```
StasDash, line 59: Create ChatWindow once, not every time you need to login

Client/Client.Qt/Application.cpp
```diff
@@ -26,12 +26,12 @@ void Application::create()
     _icon = new Style::icon(":/images/logo.png");
     _mainwidget->setBioButtonIcon(_icon);
 
+    ConnectionManager::connect();
+    std::thread(&ConnectionManager::loop).detach();
+
     _mainwidget->addWidget(std::make_unique<Login>());
     _mainwidget->addWidget(std::make_unique<Registration>());
-    _mainwidget->addWidget(std::make_unique<ChatWindow>());
```
StasDash, line 31: Get it back

StasDash (changes requested):

Client/Client.Qt/Application.cpp
GuskoMaksim: We cannot create a chat window for a user who has not logged in to the application

StasDash: We can

Client/Client.Qt/Widgets/ChannelListWindow.cpp
f1nal3: @Orange-hanter  You should point to the rule that you think was ignored.

Client/Client.Qt/Application.cpp
Orange-hanter: @StasDash give please an example of situation when we create a chat window before or authorization?

StasDash: Caching

Client/Client.Qt/Widgets/ChannelListWindow.cpp
Orange-hanter: Chapter "Formatting" rule 22

Client/Client.Qt/Application.cpp
Orange-hanter: The situation didn't clear for me. Why we should allow to create a window before authorization? 
As an example, when user manually log out from system, we have a chance to get a data from cash?

StasDash: > When user manually log out from system, we have a chance to get a data from cache? 
Depends on cache mechanism.
Also when user log out from app(app has been not closed), chat window will not be destroyed and after relogin he will use first form created.

Client/Client.Qt/Application.cpp
```diff
@@ -56,7 +56,8 @@ void Application::setAppState(App::AppState app_state)
         }
         case App::AppState::ChatWindowForm:
         {
-            _mainwidget->addWidget(std::make_unique<ChatWindow>());
+            if (_mainwidget->getCountWidgets() < 3)
```
Svidunovich, line 59: What is it for?

Client/Client.Qt/MainWidget.cpp
```diff
@@ -305,6 +305,8 @@ void MainWidget::setCentralWidget(std::int32_t index)
 
 void MainWidget::refreshTitleBar(bool showBioButton) { _title->showBioButton(showBioButton); }
 
+size_t MainWidget::getCountWidgets() { return _widgets.size(); }
```
StasDash, line 308: getWidgetsCount*

Client/Client.Qt/Application.cpp
GuskoMaksim: The chat window may contain requests related to an authorized user, but the user is not yet authorized. We can't create a chat if the user is not logged in. We already had a problem with the fact that we received channels before the user logs in and such errors will appear.

Client/Client.Qt/Application.cpp
GuskoMaksim: This is necessary to create only one chat window

Client/Client.Qt/Application.cpp
StasDash: > The chat window may contain requests related to an authorized user, but the user is not yet authorized.

After log out (w/o app close) it will.

> We can't create a chat if the user is not logged in.

We can create empty chat. 

> We already had a problem with the fact that we received channels before the user logs in and such errors will appear.

Will be fixed in https://github.com/f1nal3/Juniorgram/issues/210

Orange-hanter: @StasDash 
> chat window will not be destroyed 
Like for me, this is a bad idea. If user log out, this mean we should actually get out, and didn't keep any information in memory.
> and after relogin he will use first form created.
As you said before, "Depends on cache mechanism". We have a serialization mechanism, so we could use it as an example, storing encrypted data in the database. This is safer + faster. 
BUT. 
If user log out, we shouldn't keep any data in active memory.

Client/Client.Qt/MainWidget.cpp
```diff
@@ -305,6 +305,8 @@ void MainWidget::setCentralWidget(std::int32_t index)
 
 void MainWidget::refreshTitleBar(bool showBioButton) { _title->showBioButton(showBioButton); }
 
+size_t MainWidget::getWidgetsCount() { return _widgets.size(); }
```
fisyamaloy, line 308: "Count" is verb, say Amount

GuskoMaksim:
The request to the server and getting the list of channels temporarily remains in the constructor, until the bug with the application crash is resolved when the channels are requested again

Client/Client.Qt/Application.cpp
```diff
@@ -26,12 +26,13 @@ void Application::create()
     _icon = new Style::icon(":/images/logo.png");
     _mainwidget->setBioButtonIcon(_icon);
 
+    ConnectionManager::connect();
+    std::thread(&ConnectionManager::loop).detach();
```
StasDash, line 30: Maybe better after all widget are created?

GuskoMaksim: ok

StasDash (approved):

fisyamaloy (approved):

Orange-hanter (approved):

Issue #329, opened 2021-07-30
Successful registration should bring back to login menu


f1nal3:
Currently when you register you get a console notification of a successful registration but nothing happens on the UI.
Apart from the need to notify user via #307 we also need to bring him back to the login screen.

PR #331, opened 2021-08-02
Registration returns to login

dbski:


Svidunovich (approved):

GuskoMaksim (approved):

vDrWool (approved):

Client/Client.Qt/registration.cpp
```diff
@@ -68,6 +68,8 @@ Registration::Registration(QWidget* parent) : QWidget(parent)
         }
 
         ConnectionManager::getClient().userRegistration(email, login, password);
+
+        oApp->setAppState(App::AppState::LoginForm);
```
StasDash, line 72: What if server sends a failed result?

StasDash: I mean the name of task https://github.com/f1nal3/Juniorgram/issues/329 is on **Successful registration** but something can happen so it is not.

PR #341, opened 2021-08-07
Client connection rework

StasDash:


Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -22,149 +24,286 @@ bool Client::connect(const std::string_view& host, const uint16_t port)
...
+            default:
+                std::cerr << "[Client][Warning] unimplemented[" << uint32_t(message.mHeader.mMessageType) << "]\n";
+        }
+    }
+}
+
+void Client::onLoginAnswer(bool success)
+{
+    (void)(success);
+    std::cerr << "[Client][Warning] login answer is not implemented\n";
```
vDrWool, line 258: In such case, you should use Utility::exceptions 'NotImplementedException'.

vDrWool (changes requested):

Client/Client.Core/Client.Core.Static/Client.cpp
StasDash: Unimplemented handlers is not a critical error to use exceptions

Svidunovich (approved):

vDrWool (approved):

PR #342, opened 2021-08-07
Fix ChannelListWindow

GuskoMaksim:


Client/Client.Qt/Widgets/ChannelListWindow.cpp
```diff
@@ -100,10 +78,16 @@ void ChannelListWindow::addChannelToMainChannelWidget()
             _widgetChannelList->addItem(_channelList->takeItem(_channelList->currentRow()));
         }
     }
-    hide();
 }
 
-void ChannelListWindow::setChannels(std::vector<Network::ChannelInfo>&& channels_) { channels = channels_; }
+void ChannelListWindow::setChannels(std::vector<Network::ChannelInfo>&& channels_)
```
vDrWool, line 83: "channels_" is strange naming a bit for func arg. I guess it needs to change.

vDrWool (changes requested):

vDrWool (approved):

StasDash (approved):

PR #345, opened 2021-08-11
Implement Settings widget

### LINKED ISSUE #326: Settings widget
Next step after #84 
Implement "Setting widget". This widget is make possible for user in easy way to configure visual environment of client.
Make it ease to expand for possible features.
For now it should configure font size.

otratsiuk:
closes #326 
Currently settings are applied for TextEdit only.
After clicking to "Apply" button in Settings section, application should be restarted to display settings changes correctly. Automatically it's not implemented yet.
settings_widget

### DIFF
--- a/Client/Client.Qt/Application.cpp
+++ b/Client/Client.Qt/Application.cpp
@@ -13,8 +13,7 @@ void Application::create()
     setApplicationName("Juniorgram");
 
     Style::internal::StartFonts();
-    _mainWidget     = std::make_unique<MainWidget>();
-    _settingsWidget = std::make_unique<SettingsWidget>();
+    _mainWidget = std::make_unique<MainWidget>();
 
 #if _WIN32
     HWND handle = reinterpret_cast<HWND>(_mainWidget->winId());
@@ -71,12 +70,6 @@ void Application::setAppState(App::AppState app_state)
             _mainWidget->refreshTitleBar(true);
             _mainWidget->setCentralWidget(2);
             break;
-        }
-        case App::AppState::SettingsWidgetForm:
-        {
-            _mainWidget->refreshTitleBar(true);
-            _settingsWidget->show();
-            break;    
         }
     }
 }
--- a/Client/Client.Qt/Application.cpp
+++ b/Client/Client.Qt/Application.cpp
@@ -13,7 +13,8 @@ void Application::create()
     setApplicationName("Juniorgram");
 
     Style::internal::StartFonts();
-    _mainWidget = std::make_unique<MainWidget>();
+    _mainWidget     = std::make_unique<MainWidget>();
+    _settingsWidget = std::make_unique<SettingsWidget>();
 
 #if _WIN32
     HWND handle = reinterpret_cast<HWND>(_mainWidget->winId());
@@ -70,6 +71,12 @@ void Application::setAppState(App::AppState app_state)
             _mainWidget->refreshTitleBar(true);
             _mainWidget->setCentralWidget(2);
             break;
+        }
+        case App::AppState::SettingsWidgetForm:
+        {
+            _mainWidget->refreshTitleBar(true);
+            _settingsWidget->show();
+            break;    
         }
     }
 }
--- a/Client/Client.Qt/Application.hpp
+++ b/Client/Client.Qt/Application.hpp
@@ -3,6 +3,7 @@
 #include <memory>
 
 #include "ConnectionManager.hpp"
+#include "SettingsWidget.hpp"
 #include "Style/Style.hpp"
 
 class MainWidget;
@@ -19,7 +20,8 @@ enum class AppState
 {
     LoginForm,
     RegistrationForm,
-    ChatWindowForm
+    ChatWindowForm,
+    SettingsWidgetForm
 };
 }  // namespace App
 
@@ -54,6 +56,7 @@ public:
 
 private:
     std::unique_ptr<MainWidget>        _mainWidget;
+    std::unique_ptr<SettingsWidget>    _settingsWidget;
     std::unique_ptr<ConnectionManager> _connectionManager;
     std::unique_ptr<ReceiverManager>   _recieverManager;
     App::AppState                      _appState;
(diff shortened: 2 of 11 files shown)


Client/Client.Qt/Application.hpp
```diff
@@ -19,7 +20,8 @@ enum class AppState
 {
     LoginForm,
     RegistrationForm,
-    ChatWindowForm
+    ChatWindowForm,
+    SettingsWidgetForm
```
StasDash, line 24: First things first it's not a form

Client/Client.Qt/Application.cpp
```diff
@@ -71,6 +72,12 @@ void Application::setAppState(App::AppState app_state)
             _mainWidget->setCentralWidget(2);
             break;
         }
+        case App::AppState::SettingsWidgetForm:
+        {
+            _mainWidget->refreshTitleBar(true);
```
StasDash, line 77: It should be added to biobutton not application state

StasDash (changes requested):

Client/Client.Qt/Widgets/TitleWidget.cpp
```diff
@@ -186,6 +186,7 @@ BioButton::BioButton(QWidget* parent, bool) : CaptionButton(parent)
         menu->addAction("Username: Add format here WWWWWWWWWWWWWWWWWWWWWWWWWWW", []() {});
         menu->addSeparator();
         menu->addAction("Quit", []() { oApp->setAppState(App::AppState::LoginForm); });
+        menu->addAction("Settings", [](){ oApp->setAppState(App::AppState::SettingsWidgetForm); });
```
StasDash, line 189: also put it before separator

StasDash (changes requested):

Client/Client.Qt/Application.hpp
otratsiuk: Could you please elaborate a bit more why it should be named differently, and how do you suggest to call it instead?

StasDash: Form is a thing that switches in main window

Client/Client.Qt/Settings.cpp
```diff
@@ -19,17 +19,23 @@ Settings& Settings::getInstance()
...
+{
+    settings->clear();
+}
+
 void Settings::setFontSize(int size)
 {
     _fontSize = size;
 }
 
 int Settings::getFontSize()
```
Orange-hanter, line 36: It is a good segment of code where you could try std::optional

StasDash: Also font size mut be >0

Client/Client.Qt/Settings.cpp
```diff
@@ -19,17 +19,29 @@ Settings& Settings::getInstance()
...
 }
 
-int Settings::getFontSize()
+void Settings::setFontSize(std::uint32_t size)
 {
-    return _fontSize;
+    if (size > _minFontSize) 
+        _fontSize = size;
+    else 
+        throw std::range_error("font size value must be >0");
```
StasDash, line 36: maybe just ignore?

Client/Client.Qt/Settings.hpp
```diff
@@ -31,17 +31,24 @@ class Settings : public QSettings{
...
+     * if fontSize is not set returns indicator of optional type /
+     * with uninitialized state
      */
-    int getFontSize();
+    std::optional<std::uint32_t> getFontSize();
 
 private:
-    int _fontSize;
+    std::uint32_t _fontSize;
+    const std::uint32_t _minFontSize = 0;
```
StasDash, line 53: constexpr

Orange-hanter (changes requested):

Client/Client.Qt/Settings.cpp
StasDash, line 36: just ignore if 0

otratsiuk: @StasDash I see it differently since after calling the method, it's expected to execute some stuff. If no exception is thrown and the argument doesn't suit the condition, wouldn't it seem weird that some method was called and nothing happened?
Also there is a recommendation on our wiki to use exceptions:
image
Could you please clarify why in this particular situation exception isn't a good practice?

StasDash: > @StasDash I see it differently since after calling the method, it's expected to execute some stuff. If no exception is thrown and the argument doesn't suit the condition, wouldn't it seem weird that some method was called and nothing happened?

No.

> Also there is a recommendation on our wiki to use exceptions:
> image
> Could you please clarify why in this particular situation exception isn't a good practice?

If you throw exception application will just crash since we there is no catcher. And also if you set it font size to zero(or less) qt will just set it to system default size.

otratsiuk: @StasDash If we rely on Qt, why was the input check so necessary?

StasDash: we don't rely on qt, there should be no red message

Client/Client.Qt/Settings.hpp
```diff
@@ -32,16 +32,24 @@ class Settings : public QSettings{
      */
     void writeSettings();
     /**
-     * @brief Draft method for setting font size value
+     * @brief Method for Settiings reset by clearing .config file
+     */
+    void resetSettings();
+    /**
+     * @brief Draft method for setting font size value /
+     * if input argument is <=0, size value remains as before
```
Orange-hanter, line 40: It has no sense if user can make their font invisible. Increase minimum font size to one

Orange-hanter (changes requested):

Client/Client.Qt/Settings.hpp
otratsiuk: @Orange-hanter The user can't set font size to 0 as in SetFontSize method operator "strictly greater than" is used:
image

I see that it would be better to set  _minFontSize to 1 instead of 0. But could we please merge it as it is finally since it's just a proof of concept

Orange-hanter: @otratsiuk OK. Then please specify at documentation this moment, that we compare with variable that by default are equal to some value (zero??). I didn't deeply investigate every function.

Orange-hanter (approved):

StasDash (approved):

Issue #325, opened 2021-07-29
Implement reply mechanism on client side


Orange-hanter:
This is continuing of #20. 
For frontend based on Qt, it has a good idea to create a widget for reply mechanism.
1. Add to message reply button - just button win letter 'r'
2. After pushing this button, owner message should be copied in edit zone without any editable options. The only different, that message must have the "undo button" that stop reply.
3. Think how to mark next message as "reply"

Svidunovich:
@StasDash I need help, can you explain by what principle the scrolling appears on the right
![image](https://user-images.githubusercontent.com/47819364/129007935-fafa4724-9805-4c7c-a6a4-1e5cdf0f1fc6.png)
I about this place 
I think from the number of messages, but where the code understands that it needs to be checked, when I put the replay there, it does not think that a message has been added there, and this scroll does not appear, it appears only from the number of messages
![image](https://user-images.githubusercontent.com/47819364/129008409-a9da0c8b-65b3-4882-b34b-ce745ec88d7f.png)
I have reply in this place, but there is still a message above and does not want to let it go there

Svidunovich:
I fix it, thx

PR #346, opened 2021-08-11
Front/create reply widget

Svidunovich:
#325 
I made a client side replay mechanism
The principle of operation is as follows:
For each message, its own separate descriptor is created (for each QObject::connection a descriptor is created) which can transmit and create a replay from the message widget to the chat widget, after checking to find the widget when sending messages, the replay widget goes to the chat history widget where it is already displayed.

it looks like
image

### DIFF
--- a/Client/Client.Qt/Widgets/ChatHistory.cpp
+++ b/Client/Client.Qt/Widgets/ChatHistory.cpp
@@ -1,13 +1,15 @@
 #include "ChatHistory.hpp"
 
+#include <iostream>
+
 #include <QCoreApplication>
 #include <QDebug>
 #include <QPainter>
 #include <QtEvents>
 
-ChatHistory::ChatHistory(QWidget* parent) : QWidget(parent), _messageList()
+ChatHistory::ChatHistory(QWidget* parent) : QWidget(parent), _messageAndReplyList()
 {
-    _messageList.reserve(1024);
+    _messageAndReplyList.reserve(1024);
     _scrollArea  = std::make_unique<ScrollArea>(this);
     auto history = std::make_unique<QWidget>(this);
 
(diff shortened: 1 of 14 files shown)


Client/Client.Qt/Widgets/ChatHistory.cpp
```diff
@@ -17,37 +19,78 @@ ChatHistory::ChatHistory(QWidget* parent) : QWidget(parent), _messageList()
...
 
-    updateLayout();
+        updateLayout();
 
-    _scrollArea->scrollToWidget(msg);
+        _scrollArea->scrollToWidget(msg);
+    }
 
-    messageAdded();
+    emit messageAdded();
```
StasDash, line 84: is emit that necessary?

Client/Client.Qt/Widgets/ReplyMessageWidget.cpp
```diff
@@ -0,0 +1,65 @@
+#include "ReplyMessageWidget.hpp"
```
StasDash, line 1: as i said sometime ago It could be done without separate widget

Client/Client.Qt/Widgets/ChatHistory.cpp
Svidunovich: I think if we are using signals, then it is probably better to use emit. it's just easier to understand what it is

Client/Client.Qt/Widgets/ChatHistory.cpp
```diff
@@ -17,37 +17,78 @@ ChatHistory::ChatHistory(QWidget* parent) : QWidget(parent), _messageList()
     connect(_scrollArea.get(), SIGNAL(scrolled()), this, SLOT(resizeVisible()));
 }
 
-void ChatHistory::addMessage(const QString& message, quint64 utc, const QString& user)
+void ChatHistory::addMessage(const QString& message, quint64 utc, ReplyWidget* reply, const QString& user)
```
vDrWool, line 20: If you can try to to split this method into several methods. It's too big. (If you can't or don't want just ignore my comments)

Client/Client.Qt/Widgets/ChatHistory.cpp
```diff
@@ -68,103 +109,206 @@ void ChatHistory::resizeEvent(QResizeEvent* event)
 
 void ChatHistory::updateLayout(bool beenResized)
```
vDrWool, line 110: .

Client/Client.Qt/Widgets/ChatHistory.cpp
```diff
@@ -68,103 +109,206 @@ void ChatHistory::resizeEvent(QResizeEvent* event)
...
+            _scrollArea->scrollToY(_messageAndReplyList.back().second->pos().y());
+        }
+        else
+        {
+            _scrollArea->scrollToY(_messageAndReplyList.back().first->pos().y());
+        }
     }
 }
 
 void ChatHistory::resizeVisible()
```
vDrWool, line 192: .

Client/Client.Qt/Widgets/ChatHistory.cpp
```diff
@@ -68,103 +109,206 @@ void ChatHistory::resizeEvent(QResizeEvent* event)
...
+            resized |= bHeight || bWidth;
+            if (bWidth || bHeight) msg.second->resize(width, msg.second->expectedHeight());
+        }
     }
-    if (left >= 0 && left + 1 < int(_messageList.size())) _left = left + 1;
+    if (left >= 0 && left + 1 < int(_messageAndReplyList.size())) _left = left + 1;
     updateLayout(resized);
 }
 
 std::pair<int, int> ChatHistory::findVisible() const
```
vDrWool, line 229: .

vDrWool (approved):

PR #410, opened 2021-09-21
[Front] Channel subscription

GuskoMaksim:
1. Request for a subscription to the channel on the client side. 
2. Correction of documentation

### DIFF
--- a/Client/Client.Qt/Pages/ChannelListPage.cpp
+++ b/Client/Client.Qt/Pages/ChannelListPage.cpp
@@ -50,7 +50,7 @@ void ChannelListPage::addChannelToMainChannelWidget()
     {
         std::string channel = _channelList->currentItem()->text().toStdString();
         auto        channelInfo =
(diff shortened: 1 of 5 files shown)


Client/Client.Qt/Pages/ChannelListPage.cpp
```diff
@@ -48,25 +48,15 @@ void ChannelListPage::addChannelToMainChannelWidget()
 {
     if (_channelList->currentItem())
     {
-        if (_widgetChannelList->count() != 0)
+        std::string channel = _channelList->currentItem()->text().toStdString();
+        auto        channelInfo =
+            std::find_if(channels.begin(), channels.end(), [&channel](const Network::ChannelInfo& i)
```
vDrWool, line 53: I guess, if you don't change the var in lambda that you should pass it by value ([&channel] -> [channel]).

vDrWool (approved):

otratsiuk (approved):

PR #412, opened 2021-09-23
Reactions and time zones support

dbski:
Typo fixes

Full time zone support

Reaction storage changes

Documentation added

Schema update

Post-rebase fixes


Client/Client.Qt/Widgets/MessageWidget.cpp
```diff
@@ -18,12 +18,12 @@ MessageWidget::MessageWidget(QWidget* history, QString message, uint64_t userId,
       _messageId(messageId),
       _messageText(std::move(message)),
       _username(std::move(username)),
-      _datetime(QDateTime::fromMSecsSinceEpoch(utc)),
+      _datetime(QDateTime::fromSecsSinceEpoch(utc)),
```
StasDash, line 21: Try to use msecs since using secs might create problem in future

dbski: Could you elaborate pls?

PR #428, opened 2021-10-13
Message editing on server side

### LINKED ISSUE #378: Implement message editing on server side.
As a user I want to have an ability to edit the message that I sent earlier.

Server process looks like following:
* After receiving an edited message the server stores the updated value
* After that server responds to the client that sent this message with a status(OK if storing went good, not OK if not obviously)
* In case of a OK status server sends updates for users subscribed to channel

AC: Unit Test(if Client.CLI is done before that) code review otherwise

otratsiuk:
closes #378


Network/Public/Include/Network/Primitives.hpp
```diff
@@ -214,6 +214,41 @@ void serialize(Archive& ar, Network::MessageInfo& o)
...
+/**
+ * @brief The EditMessageInfo struct
+ * @details Edit Message Info contains sender ID, original message, edited message and message ID.
+ */
+struct EditMessageInfo
+{
+    /// sender ID uint64_t variable
+    std::uint64_t senderID;
+    /// original message string variable
+    std::string   originMessage;
```
f1nal3, line 226: Why do we need the original message?

f1nal3: Oh right, nevermind, this was discussed at the meeting.

f1nal3:
Conflict + that thing we discussed on the meeting about checking if message was edited by postgre.
Looks good otherwise =)

f1nal3 (approved):

DataAccess/DataAccess.Postgre/PostgreRepositories.hpp
```diff
@@ -44,6 +44,7 @@ namespace DataAccess
         virtual std::vector<Network::MessageInfo> getMessageHistory(const std::uint64_t channelID) override final;
         virtual Utility::StoringMessageCodes      storeMessage(const Network::MessageInfo& mi) override final;
         virtual Utility::DeletingMessageCodes     deleteMessage(const Network::MessageInfo& mi) override final;
+        virtual Utility::EditingMessageCodes      editMessage(const Network::EditMessageInfo& emi) override final;
```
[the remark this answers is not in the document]
f1nal3: I mean, if you wanna leave just one - override is the best candidate.
General rule can look like this: use virtual in the base class. Use override in the derived. Using override is guaranteeing compile check for existence  of such a function in the base class.

PR #437, opened 2021-10-18
Fix Client CMakeLists.txt

GuskoMaksim:
Bug fix
image


GuskoMaksim:
> Do you added qt path to system environment?

Yes, but Qt5Sqld.dll not enough to run the client

PR #505, opened 2022-03-30
Dev Enviroment improvement

Orange-hanter:
* CMakelists.txt wad devided to modules
* Added Ccache module
* Updated .gitignore for procecing idea files
* Import improvement of Logger module
* Some type error fix


CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.17.0)
 
+set(PROJECT "Juniorgram")
```
f1nal3: Why do we need this variable?

Base/Logger/Logger.Static/FileLogger.hpp
```diff
@@ -2,22 +2,16 @@
 
 #include <Logger/ILogger.hpp>
 #include <Utility/Utility.hpp>
-#include <algorithm>
-#include <chrono>
-#include <condition_variable>
```
f1nal3, line 7: As I've already stated - this header actually uses std::comdition_variable hence the CI issues. I won't check other headers since I believe you can easily do this using the CI.

cmake/CMakeConfig.cmake
```diff
@@ -0,0 +1,2 @@
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+set(CMAKE_VERBOSE_MAKEFILE OFF)
```
f1nal3, line 2: newline at the EOF <3
Also, project() defaults CMAKE_VERBOSE_MAKEFILE to FALSE as per https://cmake.org/cmake/help/latest/variable/CMAKE_VERBOSE_MAKEFILE.html

Base/Logger/Logger.Static/FileLogger.hpp
f1nal3, line 7: As we discussed IRL, this header actually uses std::condition_variable. Other headers can be checked using CI.

CMakeLists.txt
f1nal3: Why do we need this variable? It is never used in your PR.

cmake/ConanModule.cmake
```diff
@@ -0,0 +1,11 @@
...
+  message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+  file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/master/conan.cmake"
+                 "${CMAKE_BINARY_DIR}/conan.cmake" TLS_VERIFY ON)
+endif()
+
+include(${CMAKE_BINARY_DIR}/conan.cmake)
+
+conan_cmake_run(CONANFILE conanfile.txt
+                BASIC_SETUP CMAKE_TARGETS
+                BUILD missing)
```
f1nal3, line 11: newline at the EOF <3
But I'll be changing this part so don't bother

cmake/Warnings.cmake
```diff
@@ -0,0 +1,161 @@
...
+    set(GCC_WARNINGS
+        ${CLANG_WARNINGS}
+        -Wmisleading-indentation # warn if indentation implies blocks where blocks do not exist
+        -Wduplicated-cond # warn if if / else chain has duplicated conditions
+        -Wduplicated-branches # warn if if / else branches have duplicated code
+        -Wlogical-op # warn about logical operations being used where bitwise were probably wanted
+        -Wuseless-cast # warn if you perform a cast to the same type
+    )
+
+    if(${target_set_warnings_AS_ERROR})
```
f1nal3: This isn't used anywhere. Treating warnings as errors should be default behavior. 
Also, the naming is inconsistent: either use all caps or no caps.

f1nal3: This same variable is used in the  next set_warnings() function and it's not used yet I guess. Needs discussion.

f1nal3 (changes requested):

f1nal3 (approved):

f1nal3 (approved):

f1nal3 (approved):

f1nal3 (approved):

Agrifon (approved):

PR #526, opened 2022-04-20
Investigate the cause of a crash after a failed registration

memorisecodead:
## Changes:
1. Gone infinite loading after disconnecting from the server.
2. Added function to reconnect to the server.

## What's incomplete: 
1. Small spinner widget after connection loss, so that it would be possible to track and monitor the break with the server (Chat page).
P.S. On other pages reconnection will be held in a hidden "mode".
2. Logging on connection break and reconnection to the server in the client and server.
3. Reconnection timer.


Orange-hanter:
Please remove all useless changes like tabs and move lines

Orange-hanter:
Also, please read this article 
https://github.blog/2019-02-14-introducing-draft-pull-requests/

PR #536, opened 2022-05-05
Fixing some smells 

### LINKED ISSUE #479: Change date representation in database
## Description:
In the message header we store timestamp as std::int64_t while we use postgres's timestamp type to store it in the database.
This leads to problems while reading those values from the database due to the conversion. We should either change the way we store timestamp in database or in message header to avoid that conversion.

### Related tasks:
#478 
#476 

## Acceptance Criteria:
There is no need for additional conversions when dealing with time stored in database.
The bug described in # is not reproduced.

memorisecodead:
## Corrections:

1. Fixing unused assignments 
2. Fixing local variables

### DIFF
--- a/Client/Client.Core/Client.Core.Static/Client.cpp
+++ b/Client/Client.Core/Client.Core.Static/Client.cpp
@@ -321,7 +321,7 @@ void Client::loop()
 
             case MessageType::ServerMessage:
             {
-                /// @todo add handling 
+                // TODO add handling 
                 uint64_t clientID = 0;
                 onServerMessage(clientID);
             }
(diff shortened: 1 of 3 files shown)


ivanboika (approved):

Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -322,7 +321,6 @@ void Client::loop()
 
             case MessageType::ServerMessage:
             {
-                // @todo add handling
```
Orange-hanter, line 325: where is handling?

Orange-hanter (approved):

Orange-hanter (approved):

PR #557, opened 2022-05-25
First implementation of migration script

### LINKED ISSUE #422: Automate making changes into database layout
## Description:
Sometimes we change DB layout. The process involves stopping JG services on the host, dropping the database and then re-creating in using the updated database layout SQL file.
Such a scenario is tiresome to execute and some data(like users table for example) can be left in-between the changes.

So we need a script to do the following:
1) Stop services using the database
2) Make a copy of users table
3) Update table layout
4) Return users table
5) Start up Juniorgram services

## Acceptance Criteria:
A script file stored under git.

Orange-hanter:


Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    conn = None
+    try:
+        conn = psycopg2.connect(connection_parameters())
+        conn.autocommit = True
+        cur = conn.cursor()
+        cur.execute('DROP DATABASE IF EXISTS {};'.format(PRODUCT_DATABASE_NAME))
+        conn.commit()
+        cur.execute('CREATE DATABASE {};'.format(PRODUCT_DATABASE_NAME))
+        conn.commit()
+        cur.close()
```
memorisecodead, line 67: It's a common part, you have to take it out, because it's used a lot.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    except (Exception, psycopg2.DatabaseError) as error:
+        print(error)
+    finally:
+        if conn is not None:
+            conn.close()
+
+
+def migrate_data() -> None:
```
memorisecodead, line 91: Don't we need it here? How can we be sure that the migration will be successful?

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    functions = [make_table_copy, delete_and_create, create_tables, migrate_data]
+    for i in range(len(functions)):
+        print("Step ", functions[i].__name__)
+        functions[i]()
+        print("Done!")
+
+    if args.clean:
+        print("Step ", drop_old_db.__name__)
+        drop_old_db()
+        print("Done!")
```
memorisecodead, line 143: Enter after EOF

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/python
+# coding: utf-8
+
```
memorisecodead, line 3: Different formatting styles for lines (.format, '%s') are used, which makes code readability worse
Readability counts. (c) The Zen of Python by Tim Peters

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+new_database_name_sql = """ALTER DATABASE {0} RENAME TO _{0};
+                """.format(PRODUCT_DATABASE_NAME)
+
+
+def connection_parameters(db_name_env='POSTGRES_DB') -> str:
+    _args = "password=%s " % os.getenv('POSTGRES_PASSWORD') + \
+           "host=%s " % os.getenv('POSTGRES_HOST') + \
+           "port=%s " % os.getenv('POSTGRES_PORT') + \
+           "dbname=%s " % os.getenv(db_name_env) + \
+           "user=%s " % os.getenv('POSTGRES_USER')
```
memorisecodead, line 25: Why do we get only the name of the base right from the environment? Why not get all the parameters, considering the fact that this is not the only use in these places

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+kill_db_processes_sql = """SELECT    pg_terminate_backend (pid)
+                        FROM    pg_stat_activity
+                        WHERE   datname = '{}';
+                    """.format(PRODUCT_DATABASE_NAME)
+
+new_database_name_sql = """ALTER DATABASE {0} RENAME TO _{0};
+                """.format(PRODUCT_DATABASE_NAME)
+
+
+def connection_parameters(db_name_env='POSTGRES_DB') -> str:
```
memorisecodead, line 20: The default value for the argument makes no sense because getenv gets the variable exactly from the environment, having a local variable does not affect anything

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    _args = "password=%s " % os.getenv('POSTGRES_PASSWORD') + \
```
memorisecodead, line 21: single underscore is used for internal variables/functions (i.e., they should not be called somewhere outside the class/file). The _args variable is local and the underscore at the beginning of the name is unnecessary.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+           "user=%s " % os.getenv('POSTGRES_USER')
+    return _args
+
+
+def grab_sql_request() -> str:
+    path = "./"
+    if os.path.isfile('Database_Scheme.sql'):
+        path += 'Database_Scheme.sql'
+    else:
+        raise RuntimeError('Failed to open Database_Scheme.sql')
```
memorisecodead, line 34: If I understand the logic correctly, you want to check if there is a file with the database. You can simplify the check with os.path.exists(path_to_file) mutode and add file name to open function arguments. Or even simpler, when file doesn't exist, open function throws FileNotFoundError exception.
So function 29-38 can be rewritten in general like this
try:
    with open"./Database_Scheme.sql", "r") as source:
        request = source.read()
    return request
except FileNotFound:
    raise raise RuntimeError("File Database_Scheme.sql was not found.")

memorisecodead, line 34: Absolutely uninformative error text.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    request = None
```
memorisecodead, line 35: What is this for anyway? Even if the file is empty, request will be an empty string, not None.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    return request
+
+
+def make_table_copy() -> None:
+    conn = None
+    try:
+        conn = psycopg2.connect(connection_parameters())
+        cur = conn.cursor()
+        cur.execute(kill_db_processes_sql)
+        cur.execute(new_database_name_sql)
```
memorisecodead, line 47: I don't pretend to be an expert here, but I think you need to commit every operation with the database.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+        conn.commit()
+        cur.close()
+    except (Exception, psycopg2.DatabaseError) as error:
+        print(error)
+    finally:
+        if conn is not None:
+            conn.close()
+
+
+def delete_and_create() -> None:
```
memorisecodead, line 57: Should be renamed to something like recreate or something like that

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    conn = None
+    try:
+        conn = psycopg2.connect(connection_parameters())
+        conn.autocommit = True
```
memorisecodead, line 61: 61-64-66 Do you use both autocommit and manual commit? Why? To be sure?

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+        cur = conn.cursor()
+        cur.execute('DROP DATABASE IF EXISTS {};'.format(PRODUCT_DATABASE_NAME))
+        conn.commit()
+        cur.execute('CREATE DATABASE {};'.format(PRODUCT_DATABASE_NAME))
+        conn.commit()
```
memorisecodead, line 66: 63-66 
You should put these two operations in separate functions (and here the names of these functions can just start with an underscore), and call these two functions in this place rather than just calling the command, especially considering the fact that they are not only used here, at least delete, which means that theoretically they may still be needed in the future.

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+        print(error)
+    finally:
+        if conn is not None:
+            conn.close()
+
+
+def create_tables() -> None:
+    conn = None
+    try:
+        conn = psycopg2.connect(connection_parameters('DATABASE_NAME'))
```
memorisecodead, line 78: I assume you meant the os.getenv('DATABASE_NAME') parameter, which is the value of the PRODUCT_DATABASE_NAME variable

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def drop_old_db() -> None:
+    conn = None
+    try:
+        conn = psycopg2.connect(connection_parameters())
+        conn.autocommit = True
+        cur = conn.cursor()
+        cur.execute(kill_db_processes_sql)
+        cur.execute('DROP DATABASE IF EXISTS _{};'.format(PRODUCT_DATABASE_NAME))
```
memorisecodead, line 107: Again, either put it in a separate global query text, or put it in a prothed method (with one underscore at the beginning)

Utility/SQL/Migration.py
```diff
@@ -0,0 +1,143 @@
...
+    return _args
+
+
+if __name__ == '__main__':
+    args = parse_arguments()
+    functions = [make_table_copy, delete_and_create, create_tables, migrate_data]
+    for i in range(len(functions)):
+        print("Step ", functions[i].__name__)
+        functions[i]()
+        print("Done!")
```
memorisecodead, line 138: ITERATORS IN PYTHON FOR WHOM, ALO
From this one, please
    for i in range(len(functions))
        print("Step ", functions[i].name)
        functions[i]()
        print("Done!")

Do this:
    for function in functions:
        print("Step ", function.name)
        function()
        print("Done!")
---
One more hint: a variable of the form '_{}'.format(PRODUCT_DATABASE_NAME) can also be stored as global

Utility/SQL/Migration.py
Orange-hanter: Default value of the environment variable name, it can change.

Utility/SQL/Migration.py
Orange-hanter: For now, this is a draft. But thanks for the advice nonetheless. I would also advise you to study C++ more thoroughly instead of python zen

Utility/SQL/Migration.py
Orange-hanter: Because I'm a C++ specialist, not a Python, and the habit of working with scopes has passed from language to language. Thank you for taking the time to thoroughly search for my mistakes

Utility/SQL/Migration.py
Orange-hanter: Indeed. Since I was only planning to interrupt the flow of the script. And in case you forgot, it's a draft. Thank you for thoroughly searching for my mistakes

Utility/SQL/Migration.py
Orange-hanter: Indeed! You are not at expert here!
https://www.psycopg.org/docs/usage.html

Utility/SQL/Migration.py
Orange-hanter: You should be familiar with modular verbs in English, and feel when you are giving a strict recommendation and when you are sharing your opinion.
What about the name, I didn't bother with it since it's a draft...

memorisecodead: Daniel, calm down, this is just my opinion.

Utility/SQL/Migration.py
memorisecodead: This is just my opinion :)

Utility/SQL/Migration.py
memorisecodead: Yeaappp :)

Utility/SQL/Migration.py
memorisecodead: We are a team, we must help each other, thank you!

PR #567, opened 2022-06-15
Create new Argparser and Update Server database options

### LINKED ISSUE #504: Improve Server: expand database configuration options
## Parent Task: 
#54
#359

## Description:
For now, server configuring database connection only with **production db**. This means that the test server (which is updated with every commit to dev) also tries to connect to the production database. This affects some problems.
Look at file Server.cpp line 372 and PostgreAdapter.cpp line 11

## Acceptance Criteria:
Server is able to configure the connection to db.

KanstantsinTaukach:


DataAccess/DataAccess.Postgre/CMakeLists.txt
```diff
@@ -19,4 +18,4 @@ set_target_properties(${TARGET} PROPERTIES FOLDER ${FOLDER})
 
 target_link_libraries(${TARGET}
         PUBLIC PostgreSQL::pq libpqxx::libpqxx
-        PUBLIC Utility.Public DataAccess.Public Network.Public)
+        PUBLIC Utility.Public DataAccess.Public Network.Public Server.Core.Static)
```
Orange-hanter: Why?

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * @brief Used to take parameters from main function.
+     * @details Parses arguments which were set when application was run.
+     * @param int argc(amount of arguments), const char** argv(C-string arguments).
+     */
+    ArgParser(int argc, const char** argv) 
+    {
+        argparse::ArgumentParser program("ServerArguments");
+
+        std::cout << program << std::endl;
```
Orange-hanter: if --help

Server/Server.Core/Server.Core.Static/Server.hpp
```diff
@@ -67,9 +67,9 @@ class Server
 
 public:
     /**
-    * @brief Explicit constructor
-    * @param const uint16_t& port
-    * @details Uses incoming port value from ArgumentParser.
```
Orange-hanter, line 72: You do not have any business with third party extension here

Server/Server.Exec/MainServer.cpp
```diff
@@ -1,36 +1,17 @@
...
-        FileLogger::getInstance().log
-        (
-            e.what(), 
-            LogLevel::ERR
-        );
-    }
-
-    Server::Server server(port);
+    ArgParser parser(argc, argv);
+    Server::Server server(ArgParser::getPort());
```
Orange-hanter: Use pattern "Builder"

memorisecodead (approved):

ivanboika (commented):
Waiting for release version of pr, everything looks cool

Server/Server.Exec/MainServer.cpp
```diff
@@ -1,36 +1,28 @@
...
-    {
-        FileLogger::getInstance().log
-        (
-            e.what(), 
-            LogLevel::ERR
-        );
-    }
+    ArgParser parser(argc, argv);
+    auto      server = ServerBuilder()
+                                      .setValue(parser.getPair<std::string>("--port"))
```
Orange-hanter: Which one port, database or server?

Server/Server.Exec/CMakeLists.txt
```diff
@@ -2,7 +2,10 @@ set(TARGET Server.Exec)
 set(OUTPUT_NAME ${PROJECT}.Server.Exec)
 
 set(SOURCE_FILES
-    MainServer.cpp)
+    MainServer.cpp
+    ArgParser.hpp
+
```
Orange-hanter: Review please all code and make it more pretty, please!

Orange-hanter (changes requested):

Network/Public/Include/Network/iAPI.hpp
```diff
@@ -0,0 +1,184 @@
...
+#include "Network/SafeQueue.hpp"
+
+using ClientConn = const std::shared_ptr<Network::Connection>&;
+
+#define APPLY_API_METHODS ;\
+            void messageClient                  (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void messageAllClients              (std::shared_ptr<Network::Connection> exceptionConnection,const Message& message) const override;   \
+            void checkServerPing                (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void readAllMessage                 (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void channelListRequest             (std::shared_ptr<Network::Connection> clientConnection                           ) const override;   \
```
memorisecodead: Fix please this bracket

Network/Public/Include/Network/iAPI.hpp
```diff
@@ -0,0 +1,184 @@
...
+            void messageEditRequest             (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void messageReactionRequest         (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void registrationRequest            (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void loginRequest                   (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void channelLeaveRequest            (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void channelSubscribeRequest        (std::shared_ptr<Network::Connection> clientConnection,   const Message& message ) const override;   \
+            void channelSubscriptionListRequest (std::shared_ptr<Network::Connection> clientConnection                           ) const override;   \
```
memorisecodead: And there too

Server/Server.Exec/MainServer.cpp
```diff
@@ -1,36 +1,28 @@
...
-    catch (const std::runtime_error& e)
-    {
-        FileLogger::getInstance().log
-        (
-            e.what(), 
-            LogLevel::ERR
-        );
-    }
+    ArgParser parser(argc, argv);
+    auto      server = ServerBuilder()
```
memorisecodead: Please, explain why in this situation we should use one port? This port for database or server?

Server/Server.Exec/MainServer.cpp
```diff
@@ -10,13 +9,14 @@ using Server::Builder::ServerBuilder;
...
+    auto server = ServerBuilder()
+                                 .setValue(parser.getPair("--port"))
+                                 .setValue(parser.getPair("--dbport"))
+                                 .setValue(parser.getPair("--hostaddr"))
+                                 .setValue(parser.getPair("--dbname"))
+                                 .setValue(parser.getPair("--user"))
+                                 .setValue(parser.getPair("--password"))
+                                 .make();
 
     server->start();
```
memorisecodead, line 22: Please add a variable instead of true in the while loop, this will eliminate the magic number problem

Network/Public/Include/Network/iAPI.hpp
```diff
@@ -0,0 +1,184 @@
...
+            void channelLeaveRequest            (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
+            void channelSubscribeRequest        (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
+            void channelSubscriptionListRequest (std::shared_ptr<Network::Connection> clientConnection) const override;                             \
+            void channelDeleteRequest           (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
+            void channelCreateRequest           (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
+            void directMessageCreateRequest     (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
+            void defaultRequest                 () const override;
+
+
+
```
memorisecodead: remove useless spaces

Orange-hanter: No. This was added for easy reading

Network/Public/Include/Network/iAPI.hpp
Orange-hanter: What do you mean?

memorisecodead: Too much indentation from method name and sloppy inside arguments. We're fighting for the beauty of the code, aren't we?)

Server/Server.Exec/MainServer.cpp
Orange-hanter: Actually no. This code doesn't violate clean code principles. 
Don't look for mistakes out of the blue

Server/Server.Exec/MainServer.cpp
Orange-hanter: This is a draft. Work isn't finished.
Also, some details are predefined, and explicit initialization need in case of direct intentions.

Network/Public/Include/Network/iAPI.hpp
Orange-hanter: What you are looking for? This is macros that should be as much as possible clear. So in this case, we could violate by styleguide

f1nal3:
Hey, @KanstantsinTaukach ,since we're on the verge of a new release, please answer the following questions:
1) Do you still work on this PR?
2) How long will it take you to finish it?
3) Can it be reviewed and merged now?
4) How much will we lose if the just discard it?

Network/Public/Include/Network/iAPI.hpp
memorisecodead: image
If you're already doing it this way, you should move "const override" to the rest, like the form.

memorisecodead (changes requested):

Network/Public/Include/Network/CompressionHandler.hpp
```diff
@@ -10,7 +10,7 @@ class CompressionHandler : public AbstractHandler
 {
 public:
     /**
-     * @brief Method for compression of outcoming messages.
```
memorisecodead: As I understand it, the messages come, not go

Network/Public/Include/Network/Connection.hpp
```diff
@@ -65,7 +65,7 @@ class Connection : public std::enable_shared_from_this<Connection>
     /// References to the incoming queue of the parent object
     Utility::SafeQueue<Message>& mIncomingMessagesQueueLink;
     /// Queue all messages to be sent to the remote side of this connection
-    Utility::SafeQueue<Message> mOutcomingMessagesQueue;
+    Utility::SafeQueue<Message> mOutgoingMessagesQueue;
```
memorisecodead: Why exactly "outgoing"?

Network/Public/Include/Network/Connection.hpp
```diff
@@ -326,36 +318,36 @@ class Connection : public std::enable_shared_from_this<Connection>
...
-    void disconnect()
+        /**
+         * @brief Method for closing connection if it is opened.
+         * @details It checks if there is a connection with smb/smth. \
+         * If the connection is present, function asio::post is called, \
+         * because the current context is holding locks and \
+         * the function should be called after they have been released. This would allow \
+         * the function to acquire those locks itself without causing a deadlock.
+         */
+        void
```
memorisecodead: please make the definition look normal

Server/Server.Core/Server.Core.Static/CMakeLists.txt
```diff
@@ -4,25 +4,30 @@ set(OUTPUT_NAME Server.Core.Static)
...
-target_link_libraries(${TARGET} PUBLIC Network.Public DataAccess.Public Utility.Public Base.Logger.Static)
 
-set_target_properties(${TARGET} PROPERTIES LINKER_LANGUAGE CXX)
+
+set_target_properties(${TARGET} PROPERTIES	LINKER_LANGUAGE CXX
+											FOLDER 			${FOLDER})
+
+target_include_directories(${TARGET}	INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}
+										PRIVATE $<TARGET_PROPERTY:DataAccess.Postgre,SOURCE_DIR>/..
+										PRIVATE $<TARGET_PROPERTY:DataAccess.Public,SOURCE_DIR>/..)
```
memorisecodead: fix last line

Server/Server.Core/Server.Core.Static/Server.cpp
```diff
@@ -1,13 +1,78 @@
...
+
 
 namespace Server
 {
 using namespace DataAccess;
 using UtilityTime::safe_localtime;
 
+
+
+Server::Server() { /* do nothing*/ }
```
memorisecodead: If we want to stick to the logic of the builder server, it must have default values passed to the default constructor

Server/Server.Core/Server.Core.Static/Server.cpp
```diff
@@ -1,13 +1,78 @@
...
+
+
+void Server::initRepository(std::unique_ptr<PostgreRepositoryManager> postgreManager)
+{
+    _postgreManager.swap(postgreManager);
+}
+
```
memorisecodead: We can't just initialize these values without a constructor. These class members must be initialized first, so that we can change them later if necessary

Server/Server.Core/Server.Core.Static/Server.cpp
```diff
@@ -25,6 +90,55 @@ void Server::onClientDisconnect(const std::shared_ptr<Connection>& client)
...
+            LogLevel::INFO
+        );
+
+        auto newConnection = 
+            std::make_shared<Connection>(Connection::OwnerType::SERVER, _context, std::move(socket), _incomingMessagesQueue);
+
+        if (onClientConnect(newConnection))
+        {
+            _connectionsPointers.emplace_back(std::move(newConnection));
+            _connectionsPointers.back()->connectToClient(_idCounter++);
```
memorisecodead, line 89: Do we need a connection with a null user?

Server/Server.Core/Server.Core.Static/Server.cpp
```diff
@@ -145,131 +259,50 @@ void Server::onMessage(const std::shared_ptr<Connection>& client, Message& messa
...
     {
-        if (thread.joinable())
-        {
-            thread.join();
-        }
-    }
+        onClientDisconnect(client);
 
-    _threads.clear();
+       // client.reset();
```
memorisecodead: I think the thing to do here is to send a message about the wrong message, and so collect logs without putting the wrong message in the container.

Server/Server.Core/Server.Core.Static/Server.cpp
```diff
@@ -145,131 +259,50 @@ void Server::onMessage(const std::shared_ptr<Connection>& client, Message& messa
...
 
-        _connectionsPointers.erase(std::remove(_connectionsPointers.begin(), _connectionsPointers.end(), client),
-                                   _connectionsPointers.end());
+        onMessage(messageInfo.mRemote, messageInfo);
+        messagesCount++;
     }
 }
 
-void Server::messageAllClients(const Message& message, const std::shared_ptr<Connection>& exceptionClient)
+void Server::messageAllClients(std::shared_ptr<Connection> exceptionClient, const Message& message) const
```
memorisecodead, line 269: Why did you add a message type variable to the method arguments?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,60 @@
...
+        ServerBuilder() = default;
+        ~ServerBuilder() = default;
+ 
+        inline ServerBuilder& setValue(std::pair<std::string, std::string> keyValue)
+        {
+            _arguments.insert(keyValue);
+            return *this;
+        }
+
+        Server* make()
```
memorisecodead: Are you sure that's the right title?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+            auto obj = new Server();
```
memorisecodead: Why do you use a clean pointer instead of a smart pointer?

DataAccess/Public/Include/DataAccess/AbstractRepositoryManager.hpp
```diff
@@ -8,158 +8,148 @@
...
+    std::priority_queue<RepositoryRequest>       _queue;
+
+    std::mutex              _queueMutex;
+    std::condition_variable _queueCV;
+
+    std::atomic<bool> _handlerState = true;
+    std::thread       _repositoryRequestsHandler;
+
+public:
+    IRepositoryManager() {}
```
memorisecodead: We should make this constructor the default, since we're not going to move it anywhere, as I understand, when we have the move constructor removed, as well as the copy constructor.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: Other suggestions?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: Try to fix. You will find out that because of the hidden constructor, you can't use smart pointers

Server/Server.Core/Server.Core.Static/Server.cpp
Orange-hanter: Explain in detail what you mean

Server/Server.Core/Server.Core.Static/Server.cpp
Orange-hanter: Look closely, this type has already been

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
```diff
@@ -1,6 +1,7 @@
 #include <catch2/catch.hpp>
 
 #include "DataAccess.Postgre/PostgreRepositoryContainer.hpp"
+#include "DataBaseOptions.hpp"
```
ivanboika, line 4: Why this should be included?

DataAccess/Public/Include/DataAccess/QueryBuilder.hpp
```diff
@@ -18,33 +14,33 @@ namespace DataAccess
  * @details The class forms and saves SQL query, current
  * table and type of database. Class provides interface, which is as close as possible
  * to the classical SQL queries. This is achieved by calling methods one by one.
- * Examples for postgre database you can see in PosgreQuery.hpp
+ * Examples for postgre database you can see in PostgreQuery.hpp
```
ivanboika: I guess PostgreQueryBuilder?

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
Orange-hanter: This is testing, hardcode parameters are not prohibited here

DataAccess/Public/Include/DataAccess/QueryBuilder.hpp
Orange-hanter: Yep, you are right. File should be renamed

DataAccess/DataAccess.Debug/FileRepository.cpp
```diff
@@ -1,7 +1,7 @@
 #include "FileRepository.hpp"
 
-#include "Utility/Exception.hpp"
-#include "Utility/Utility.hpp"
+#include <Utility/JGExceptions.hpp>
```
f1nal3, line 3: Why this differs from the syntax in FileLogger.cpp ?

Network/Public/Include/Network/Connection.hpp
```diff
@@ -14,21 +14,21 @@
...
 #include "YasSerializer.hpp"
 
 namespace Network
 {
-   
+
 /** @class Connection
- *  @brief Connection managment class
+ *  @brief Connection management class
  *  @details see https://github.com/f1nal3/Juniorgram/wiki/Legacy-Frontend-network \
```
f1nal3, line 32: This should be changed into Doxygen reference

Network/Public/Include/Network/Message.hpp
```diff
@@ -8,7 +8,7 @@
 #include <memory>
 #include <vector>
 
-#include "Connection.hpp"
+#include "Network/Connection.hpp"
 #include <Utility/Utility.hpp>
```
f1nal3, line 12: There seems to be no consistence in "_" or <_> usage

DataAccess/DataAccess.Postgre/PostgreRepositoryManager.hpp
```diff
@@ -13,7 +14,17 @@ namespace DataAccess
 class PostgreRepositoryManager: public IRepositoryManager
 {
 public:
-	explicit PostgreRepositoryManager(const std::shared_ptr<IAdapter>& repositoryContainer)
-		: IRepositoryManager(repositoryContainer) {}
```
memorisecodead, line 17: You could take out the repository registration method and leave the constructor, thus reducing the code from 8 lines to 3.

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
memorisecodead: If you look at line 13, the namespace from this header is used.

DataAccess/Public/Include/DataAccess/AbstractRepositoryContainer.hpp
```diff
@@ -4,6 +4,7 @@
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <tuple>
```
memorisecodead: Why not declare this header where its functionality is used? (AbstractRepositoryManager.hpp)

DataAccess/Public/Include/DataAccess/FutureResult.hpp
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#include <any>
+#include <future>
+
```
memorisecodead, line 5: Remove unnecessary space

DataAccess/Public/Include/DataAccess/RepositoryRequest.hpp
```diff
@@ -94,4 +94,4 @@ namespace DataAccess
         }
     };
 
-} /// namespace DataAccesss
+} /// namespace DataAccess
```
memorisecodead, line 97: Remove one extra "/"

Network/Public/Include/Network/Connection.hpp
```diff
@@ -325,36 +317,33 @@ class Connection : public std::enable_shared_from_this<Connection>
         }
     }
 
-    suppressWarning(4100, "-Wunused-parameter")
```
memorisecodead, line 328: Why did you remove the macro to ignore the error?

Server/Server.Core/Server.Core.Static/Server.cpp
memorisecodead: > Do we need a connection with a null user?

Server/Server.Core/Server.Core.Static/Server.cpp
memorisecodead: > Why did you add a message type variable to the method arguments?

Server/Server.Core/Server.Core.Static/Server.hpp
```diff
@@ -20,230 +20,120 @@ class PostgreRepositoryManager;
...
 
     /**
-     * @brief The method for checking the registration request.
-     * @param const std::shared_ptr<Connection>& client - for connect messages to the client or server.
-     * @param Message& message - for use the message header functionality.
+     * @brief Default constructor.
+     * @details Initialize Server object. After you steel need to initialize network and database connection.
      */
-    void registrationRequest(const std::shared_ptr<Connection>& client, Message& message) const;
+    Server() = default;
```
memorisecodead, line 90: Shouldn't the server have a different builder without taking the builder server into account?
And if you need to run a server without a builder?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <cstdint>
+#include <map>
+#include <memory>
+#include <string>
+
+#include "Server.hpp"
+
+namespace Server::Builder
+{
+
```
memorisecodead: Remove unnecessary space

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+    {
+        std::unique_ptr<Server> uptr(makeServer());
+        return uptr;
+    }
+
+private:
+    std::map<std::string, std::string>  _arguments;
+    RepoManager_uptr                    _repository;
+};
+}  // namespace Server::Builder
```
memorisecodead: Maybe you meant "ServerBuilder"?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+#include <FileLogger.hpp>
+
+/**
+ * @class ArgParser.
+ * @brief This class parses arguments with which application is started.
+ * @details
+ * For example: "--port 5432 --dbport 65001 --dbname juniorgram --hostaddr 127.0.0.1 --user postgres --password postgres".
+ * Sending the port to the Server class constructor and list of all arguments to the PostgreAdapter Instance Method.
+ */
+
```
memorisecodead: Remove unnecessary space

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+                {
+                    Base::Logger::FileLogger::getInstance().log("Bad port value", Base::Logger::LogLevel::ERR);
+                    
+                    return 5432;
+                }
+                return port;
+            });
+
+        _program.add_argument("--dbport")
+            .default_value("65001")
```
memorisecodead: Why not make a variable and assign it a port value?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+            .help("Set the database port value")
+            .action([](std::string value) -> int {
+                int port = std::stoi(value);
+                if (port > std::numeric_limits<uint16_t>::max())
+                {
+                    Base::Logger::FileLogger::getInstance().log("Bad DataBase port value", Base::Logger::LogLevel::ERR);
+
+                    return 65001;
```
memorisecodead: What is the "65001"? If they returned a variable, it would be clear

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+     */
+    ArgParser(int argc, const char** argv) 
+    {
+        _program = argparse::ArgumentParser("ServerArguments", "0.6", argparse::default_arguments::none);
+        
+        _program.add_description("Server side of juniorgram messenger. Just have a fun!");
+        _program.add_epilog("Some part of program could be gluchennie.");
+
+        _program.add_argument("--port")
+            .default_value("5432")
```
memorisecodead: Why not make a variable and assign it a port value?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+            .help("Set the port value for server needs")
+            .action([](std::string value) -> int {
+                int port = std::stoi(value);
+                if (port > std::numeric_limits<uint16_t>::max())
+                {
+                    Base::Logger::FileLogger::getInstance().log("Bad port value", Base::Logger::LogLevel::ERR);
+                    
+                    return 5432;
```
memorisecodead: What is the "5432"? If they returned a variable, it would be clear

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+    ArgParser(const ArgParser&) = delete;
+   
+    /**
+     * @brief Used to take parameters from main function.
+     * @details Parses arguments which were set when application was run.
+     * @param int argc(amount of arguments), const char** argv(C-string arguments).
+     */
+    ArgParser(int argc, const char** argv) 
+    {
+        _program = argparse::ArgumentParser("ServerArguments", "0.6", argparse::default_arguments::none);
```
memorisecodead: What is the "0.6"?
I know that this is the parser version...

Utility/Utility.Public/Include/Utility/JGExceptions.hpp
```diff
@@ -40,7 +40,7 @@ class JuniorgramException : public std::exception
 private:
     std::string   _msg;
     std::string   _file;
-    std::uint16_t _line;
+    [[maybe_unused]] std::uint16_t _line;
```
memorisecodead, line 43: What is the premise of [[maybe_unused]] attribute?

memorisecodead (commented):
Performed a review.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+        {
+            auto up_repo = std::make_unique<DataAccess::PostgreRepositoryManager>(IAdapter::getInstance<DataAccess::PostgreAdapter>(dbOptions));
+            obj->initRepository(std::move(up_repo));
+        }
+        else
+        {
+            obj->initRepository(std::move(_repository));
+        }
+
+        // initialize host port
```
memorisecodead: Remove usless como

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+        uint16_t port = std::stoi(_arguments["--port"]);
+        obj->initConnection(port);
+
+        return obj;
+    }
+
+    std::unique_ptr<Server> makeServerRAII() noexcept
```
memorisecodead: Why do you use noexcept here?

memorisecodead: Why is makeServerRAII made inside the ServerBuilder and not as a separate function independent of that class?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+        return *this;
+    }
+
+    inline ServerBuilder& setValue(DataAccess::IRepositoryManager* repoManager)
+    {
+        _repository.reset(repoManager);
+        return *this;
+    }
+
+    Server* makeServer() noexcept
```
memorisecodead: Why do you use noexcept here?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+        if (!_repository)
+        {
+            auto up_repo = std::make_unique<DataAccess::PostgreRepositoryManager>(IAdapter::getInstance<DataAccess::PostgreAdapter>(dbOptions));
+            obj->initRepository(std::move(up_repo));
+        }
+        else
+        {
+            obj->initRepository(std::move(_repository));
+        }
+
```
memorisecodead, line 75: Why not put the code (59-70) in a separate function?

Network/Public/Include/Network/iAPI.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+
+namespace Network{
+    class iAPI
```
memorisecodead, line 32: Which API is it? I would mark specifically what kind of API it is. (conventionally speaking ClientAPI).
On line 7 you use using ClientConn. So it's a ClientAPI?

Network/Public/Include/Network/Connection.hpp
```diff
@@ -282,8 +270,12 @@ class Connection : public std::enable_shared_from_this<Connection>
      */
     Connection(const OwnerType& owner, asio::io_context& contextLink, asio::ip::tcp::socket socket,
                Utility::SafeQueue<Message>& incomingMessagesQueueLink)
-        : mOwner(owner), mSocket(std::move(socket)), mContextLink(contextLink), _readStrand(contextLink), _writeStrand(contextLink), 
-            mIncomingMessagesQueueLink(incomingMessagesQueueLink)
+        : mOwner(owner),
+          mSocket(std::move(socket)),
+          mContextLink(contextLink),
+          _readStrand(contextLink),
+          _writeStrand(contextLink),
```
memorisecodead, line 285: CamelCase or underscore, I don't understand

Network/Public/Include/Network/Connection.hpp
```diff
@@ -50,7 +50,7 @@ class Connection : public std::enable_shared_from_this<Connection>
     OwnerType mOwner = OwnerType::SERVER;
     /// Connection id
     std::uint64_t mConnectionID = uint64_t();
-    std::uint64_t userID = 1;
+    std::uint64_t userID        = 1;
```
memorisecodead: Why is there an indentation?

DataAccess/Public/Include/DataAccess/QueryBuilder.hpp
```diff
@@ -3,11 +3,7 @@
 #include <iostream>
 #include <string>
```
memorisecodead, line 5: Please rename the file to match the class name "AbstractQueryBuilder"

DataAccess/Public/Include/DataAccess/QueryBuilder.hpp
```diff
@@ -18,33 +14,33 @@ namespace DataAccess
...
 public:
-    QueryBuilder(Utility::DatabaseType type, const std::string& tableName, std::shared_ptr<IAdapter> adapter)
-        : _databaseType{ type }, _tableName{ tableName }, _statement{ nullptr }, _adapter{ adapter } {}
+    AbstractQueryBuilder(Utility::DatabaseType type, const std::string& tableName, std::shared_ptr<IAdapter> adapter)
+        : _databaseType{type}, _tableName{tableName}, _statement{nullptr}, _adapter{adapter}
+    {
+    }
 
-    virtual ~QueryBuilder(void) { this->clearStatement(); }
+    virtual ~AbstractQueryBuilder(void) { this->clearStatement(); }
```
memorisecodead: In C++, the default in parentheses when no arguments are given is void. Please fix this

memorisecodead (commented):
A little bit more review

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: Typically, used method `make`. If you have any better option, go on, open to suggestions

Server/Server.Core/Server.Core.Static/Server.cpp
Orange-hanter: The server could be constructed using an already existing object. That's the idea of builder pattern. Other question why am I used here qualified type instead of interface type. That is wrong...

Server/Server.Core/Server.Core.Static/Server.cpp
Orange-hanter: The server could be constructed using an already existing object. That's the idea of builder pattern. Other question why I use here qualified type instead of interface type

DataAccess/DataAccess.Postgre/PostgreRepositoryManager.hpp
Orange-hanter: By and large, Data Access is a framework. A framework implies contract programming. In this case, our contract is an agreement to define the `registerRepositories()` method. If you try to transfer these activities to the constructor, then you need to change the structure of the inherited interface.
I hope you understand what the problem is, I will ask about it in one of the next sessions.

DataAccess/Public/Include/DataAccess/AbstractRepositoryContainer.hpp
Orange-hanter: I have already prepared a patch in which all these minor bugs are fixed. Waiting for Valentine's review

DataAccess/Public/Include/DataAccess/RepositoryRequest.hpp
Orange-hanter: Styleguide. Chapter "Comments" rule 8

Server/Server.Exec/ArgParser.hpp
Orange-hanter: This magic constant gets the context from the code. There is no need to go overboard with the extra variable here. We can see that this is the default value for the database port.

Server/Server.Exec/ArgParser.hpp
Orange-hanter: See above

Server/Server.Exec/ArgParser.hpp
Orange-hanter: See above

Server/Server.Exec/ArgParser.hpp
Orange-hanter: Ve haven't any versioning tool, so for now it will be like that

Utility/Utility.Public/Include/Utility/JGExceptions.hpp
Orange-hanter: goole it

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
ivanboika: Yeah, but I mean that if those hardcode params were used without this, why do we need to reinclude?

Utility/Utility.Public/Include/Utility/JGExceptions.hpp
f1nal3: @memorisecodead it's exactly the fix to this warning I mentioned to you before.

Server/Server.Core/Server.Core.Static/Server.cpp
ivanboika: Why did you delete const and & from shared_ptr? I guess there we don't change state of ownership

Server/Server.Core/Server.Core.Static/Server.hpp
ivanboika: That's a build pattern and from this point every server instance must be build by builder and that means that we maybe(not so sure) can't run server without builder

ivanboika: Two builders for 1 instance, quite strange
I don't see any advantages from this

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+                options += arg.second;
+                options += " ";
+            }
+        }
+
+        const char* dbOptions = options.c_str();
+        using DataAccess::IAdapter;
+        if (!_repository)
+        {
+            auto up_repo = std::make_unique<DataAccess::PostgreRepositoryManager>(IAdapter::getInstance<DataAccess::PostgreAdapter>(dbOptions));
```
ivanboika: Maybe I'm not right, but who do we use postgre instance in server builder? I thought our idea was to be more flexible in the way which repo we wanna use or define what repo we're using when server starts. I can understand it now because right now we have only 1 repo ready for use

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,88 @@
...
+        // initialize host port
+        uint16_t port = std::stoi(_arguments["--port"]);
+        obj->initConnection(port);
+
+        return obj;
+    }
+
+    std::unique_ptr<Server> makeServerRAII() noexcept
+    {
+        std::unique_ptr<Server> uptr(makeServer());
```
ivanboika: Why not to use make_unique?

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
memorisecodead: Reinclude, where? 
I guess, this is the .cpp with the tests. Perhaps in the future there will be other functionality that will require the inclusion of this header.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
ivanboika: MainServer.cpp
I guess it's convenient to use

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,130 @@
...
+        
+        _program.add_description("Server side of juniorgram messenger. Just have a fun!");
+        _program.add_epilog("Some part of program could be gluchennie.");
+
+        _program.add_argument("--port")
+            .default_value("5432")
+            .help("Set the port value for server needs")
+            .action([](std::string value) -> int {
+                int port = std::stoi(value);
```
ivanboika: I see restrictions only for positive numbers, but what if we have number below zero? Maybe I've lost check about this. Also, why int. I mean, everywhere when we don't need to see numbers below zero we use unsigned version and sometimes it says something about restrictions and properies of variable (like I said about that we don't look forward to seeing negative numbers)

Server/Server.Exec/ArgParser.hpp
f1nal3: Can't check it myself atm, sorry, but maybe we can get the default value for parameter and use them? Something like `_program.get("port something").default() ` ?

Server/Server.Exec/ArgParser.hpp
memorisecodead: > I see restrictions only for positive numbers, but what if we have number below zero? Maybe I've lost check about this. Also, why int. I mean, everywhere when we don't need to see numbers below zero we use unsigned version and sometimes it says something about restrictions and properies of variable (like I said about that we don't look forward to seeing negative numbers)

We need to branch for this situation, so that in case of a negative number the builder will not start or it will start on one of the default ports 65001 or 5432.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
f1nal3: I would rather suggest hiding make() by making it private, we don't really wanna hand around pointers like that.
It's probably a nitpick, but there are more ways to ensure RAII than unique_ptr, so I kinda dislike the name a little bit.

DataAccess/DataAccess.Test/PostgreRepositoriesContainerTest.cpp
ivanboika: If this file already uses stuff from .hpp, because this .hpp was included somewhere else, but this somewhere else is included there and we still can use it
I mean this situation

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: You're right. At the moment, this is some gag so that the server does not break. I commented above the code that we need a more elegant way to make this work.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: Server has private constructor and haven't copy constructor, make_unique don't work in this way

Server/Server.Core/Server.Core.Static/Server.hpp
memorisecodead: > Two builders for 1 instance, quite strange I don't see any advantages from this

Which two instances are we talking about?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
ivanboika: Yeah, I've seen this comment but thought that's not about that point

memorisecodead (approved):

Network/Public/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ set(SOURCE_FILES
     Include/Network/Message.hpp
     Include/Network/SerializationHandler.hpp
     Include/Network/YasSerializer.hpp
-    Include/Network/iAPI.hpp)
+    Include/Network/IClientAPI.hpp)
```
Orange-hanter: Decline. We have an API. It didn't divide on client or server

Network/Public/Include/Network/IClientAPI.hpp
```diff
@@ -4,14 +4,12 @@
...
 
-using ClientConn = const std::shared_ptr<Network::Connection>&;
-
 #define APPLY_API_METHODS \
             void messageClient                  (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
             void messageAllClients              (std::shared_ptr<Network::Connection> exceptionConnection,const Message& message) const override;   \
             void checkServerPing                (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
             void readAllMessage                 (std::shared_ptr<Network::Connection> clientConnection,   const Message& message) const override;   \
-            void channelListRequest             (std::shared_ptr<Network::Connection> clientConnection) const override;                             \
+            void channelListRequest             (std::shared_ptr<Network::Connection> clientConnection)   const override;                           \
```
Orange-hanter: Reject. I didn't leave this space on purpose.

Network/Public/Include/Network/IClientAPI.hpp
```diff
@@ -23,21 +21,17 @@ using ClientConn = const std::shared_ptr<Network::Connection>&;
...
 
-
-
-
-
-namespace Network{
-    class iAPI
+namespace Network
+{
+    class IClientAPI
```
Orange-hanter: Reject naming. See above

Orange-hanter (changes requested):

Network/Public/Include/Network/IClientAPI.hpp
memorisecodead: This API is implemented for whom/what?

Network/Public/CMakeLists.txt
memorisecodead: This API is implemented for whom/what?

Network/Public/Include/Network/IClientAPI.hpp
memorisecodead: If you pay attention to the methods you made for the API, they clearly scream that they are for the client. 

Explain to me who/what this API is for?

Network/Public/CMakeLists.txt
Orange-hanter: Understand what an API is first
https://en.wikipedia.org/wiki/API

memorisecodead: > Understand what an API is first https://en.wikipedia.org/wiki/API

This API is implemented for whom/what?

memorisecodead: > Understand what an API is first https://en.wikipedia.org/wiki/API

image

memorisecodead: > Understand what an API is first https://en.wikipedia.org/wiki/API

Am I correct in assuming that the API is made for "custom clients" or some other entity? If it's for a custom client, it makes sense to add "Client" prefix to the beginning of the class name, and if it's for another entity, then another prefix.

DataAccess/DataAccess.Postgre/PostgreAdapter.cpp
```diff
@@ -8,18 +8,18 @@ namespace DataAccess
 
         if (_instance == nullptr)
         {
-            _instance = std::shared_ptr<PostgreAdapter>(new PostgreAdapter(options.empty() ? _defaultOptions : options));
+            _instance = std::shared_ptr<PostgreAdapter>(new PostgreAdapter(options));
```
Orange-hanter, line 11: Better change to make_shared

DataAccess/DataAccess.Postgre/PostgreAdapter.cpp
```diff
@@ -48,10 +48,10 @@ namespace DataAccess
...
 
-    void PostgreAdapter::closeConnection(void) 
+    void PostgreAdapter::closeConnection() 
     { 
         _instance.reset();
         _connection.reset();
     }
 
-} ///namespace DataAccess
+} // namespace DataAccess
```
Orange-hanter, line 57: Move back changes, to styleguide consistency

DataAccess/DataAccess.Postgre/PostgreAdapter.cpp
memorisecodead: That would be cool, but there is a problem enclosing the members of the PostgreAdapter class. When instantiating a smart pointer to a PostgreAdapter object, we can't access the private _connection member when the PostgreAdapter is constructed. Also, _connection is of type std::unique_ptr<pqxx::connection>. 

image

f1nal3 (approved):

DataAccess/DataAccess.Postgre/PostgreAdapter.cpp
Orange-hanter: So, it has sense to make public constructor?

Server/Server.Exec/MainServer.cpp
```diff
@@ -1,36 +1,29 @@
...
+                                 .setValue(parser.getPair("--hostaddr"))
+                                 .setValue(parser.getPair("--port"))
+                                 .setValue(parser.getPair("--user"))
+                                 .setValue(parser.getPair("--password"))
+                                 /*.setValue("managerName")*/
+                                 .makeServer();
 
-    Server::Server server(port);
-    server.start();
+    serverBuild->start();
```
Orange-hanter: Server build?? Change to application\app or server

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * @class ArgParser.
+ * @brief This class parses arguments with which application is started.
+ * @details
+ * For example: "--serverport 5432 --port 65001 --dbname juniorgram --hostaddr 127.0.0.1 --user postgres --password postgres".
+ * Sending the port to the Server class constructor and list of all arguments to the PostgreAdapter Instance Method.
+ */
+class ArgParser
+{
+private:
+    argparse::ArgumentParser _argParser;
```
Orange-hanter: Why do you move params at start?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,140 @@
...
+     * @brief It's deleted copy constructor. There is no need to use this.
+     */
+    ArgParser(const ArgParser&) = delete;
+   
+    /**
+     * @brief Used to take parameters from main function.
+     * @details Parses arguments which were set when application was run.
+     * @param int argc(amount of arguments), const char** argv(C-string arguments).
+     */
+    explicit ArgParser(int argc, const char** argv) noexcept
```
Orange-hanter: Check logger behaviour in case of raise exception.

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,140 @@
...
+                if (port > std::numeric_limits<uint16_t>::max())
+                {
+                    Base::Logger::FileLogger::getInstance().log
+                    (
+                        "Bad DataBase port value",
+                        Base::Logger::LogLevel::ERR
+                    );
+                    return 5432;
+                }
+                return port;
```
Orange-hanter: Well, any issue here?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,140 @@
...
+            argparse::default_arguments::none);
+        
+        _argParser.add_description("Server side of juniorgram messenger. Just have a fun!");
+        _argParser.add_epilog("Some parts of the program may be configured by default.");
+
+        _argParser.add_argument("--serverport")
+            .default_value<std::string>("65001") 
+            .help("Set the port value for server needs")
+            .action([](const std::string& value) -> uint16_t {
+                uint16_t port = std::stoi(value);
```
Orange-hanter: Do static cast to suppress warnings (same at line 64)

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -0,0 +1,85 @@
...
+        std::unique_ptr<Server> ptr(make());
+        return ptr;
+    }
+
+private:
+    Server* make()
+    {
+        // form a string with the configuration for connecting to the database
+        std::string dbOptions = "";
+        for (const auto& [firstArg, secondArg] : _arguments)
```
Orange-hanter: "firstArg, secondArg", excuse me?

Server/Server.Exec/ArgParser.hpp
```diff
@@ -0,0 +1,140 @@
...
+    ArgParser(const ArgParser&) = delete;
+   
+    /**
+     * @brief Used to take parameters from main function.
+     * @details Parses arguments which were set when application was run.
+     * @param int argc(amount of arguments), const char** argv(C-string arguments).
+     */
+    explicit ArgParser(int argc, const char** argv) noexcept
+    {
+        _argParser = argparse::ArgumentParser("ServerArguments", "1.0",
```
Orange-hanter: Return back version

Orange-hanter (changes requested):

Orange-hanter:
Also, move ArgParser to Server.Core. Why - I will answer on Wednesday.

BEFORE MEETING WEDNESDAY DO NOT CLOSE PR

Server/Server.Exec/ArgParser.hpp
memorisecodead: Why is version 0.6 preferable to 1.0?

DataAccess/DataAccess.Postgre/PostgreAdapter.cpp
memorisecodead: By default Doxygen formats the end of the namespace with //. 
I may not have looked carefully just now, correct me, but there is no such namespace formatting in the styleguide .

Server/Server.Exec/ArgParser.hpp
memorisecodead: Excuse me?

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -30,14 +30,116 @@ class ArgParser
...
+                    (
+                        "Bad port value", 
+                        Base::Logger::LogLevel::ERR
+                    );
+                    return std::to_string(65001);
+                }
+                return std::to_string(port);
+            });
+
+        _argParser.add_argument("--port")
```
Orange-hanter, line 58: The argument should have an obvious name, so rename it to '--portDB' or something like that

Orange-hanter: Below is an example "--dbname"

Orange-hanter (changes requested):

Server/Server.Core/Server.Core.Static/ArgParser.hpp
memorisecodead: I already answered this question on Monday. 
Just a reminder: Postge is bound to the default flags (--port, --user and etc). 

Solution: use the default flags or trim the db prefix at the time of initialization.

Orange-hanter: Need to transfer this logic directly to the database adapter. let's do it in another task

Orange-hanter (approved):

PR #576, opened 2022-06-26
Fixing some details in PostgreRepositories.cpp

### LINKED ISSUE #575: Fixing some details in PostgreRepositories.cpp
## Parent Task: 
#191 

## Description:
If we gonna start to work, we have some problems with process data in postgreRepositories.
This problem with lack "equal" value in one part of postgreRepositories. Also, we have some problems with Logger part, because in work proccess, we can see strange output.

memorisecodead:


Orange-hanter (approved):
Looks good

slavexx (approved):

ivanboika (approved):

PR #582, opened 2022-07-05
Fixing bad output from Logger

### LINKED ISSUE #579: Improve logger
### Our logger eats a part of information that we can see in the terminal. We need to fix this issue 
image
### We have back in future issue
image

## Acceptance Criteria:
Logger shows us information in appropriate way, without eating parts of data

## Possible solutions:
I guess, we can find problem around threads. Maybe adding any mutex or something else will fix it. Or not
##How to produce:
You can start any type of tests (better DataAccess part) and you'll see in the terminal incorrect output
Also, we may launch client&server and check how logger shows us the output

ivanboika:


memorisecodead (approved):

Base/Logger/Logger.Static/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set_target_properties(${TARGET} PROPERTIES
 target_link_libraries(${TARGET} PUBLIC
         Base.Logger.Public
         Utility.Public
+        Network.Public
```
Orange-hanter, line 19: Why do you connect network to loggere??

Orange-hanter (changes requested):

Base/Logger/Logger.Static/CMakeLists.txt
ivanboika: For SafeQueue

Orange-hanter: Hmm, it isn't good. Better move SafeQueue to Utils

ivanboika: It will be a new task, am I right?

Orange-hanter: You could push this PR. But please, make new Issue for moving SafeQueue to Utils :)

Orange-hanter (approved):

ivanboika:
I'll wait another one review and then gonna push it

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
```diff
@@ -188,8 +188,7 @@ Utility::ChannelCreateCodes ChannelsRepository::createChannel(const Models::Chan
     _pTable->Insert()->columns(SubscribNewChannelData)->execute();
 
     Base::Logger::FileLogger::getInstance().log(
-        '['+ "channel id: " + std::to_string(IDNewChannel) +
-        "] Creating channel was successful\n",
+        std::string("Creating channel was successful\n"),
```
slavexx: I think, when we create message for logs, we should give more information about our actions. I suggest leave ID of channel or add the name of new channel in log message

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
```diff
@@ -219,7 +218,7 @@ std::uint64_t LoginRepository::loginUser(const Models::LoginInfo& loginInfo)
     catch (const std::exception& e)
     {
         Base::Logger::FileLogger::getInstance().log(
-            std::string(e.what() +'\n'),
+            std::string(std::string(e.what()) +'\n'),
```
slavexx, line 222: Why are we convert C-style string to std::string? Is it so necessary?

DataAccess/Public/Include/DataAccess/SQLStatements.hpp
```diff
@@ -243,7 +243,7 @@ class SQLBase
         catch (const std::exception& err)
         {
             Base::Logger::FileLogger::getInstance().log(
-                std::string(err.what() + '\n'),
+                std::string(std::string(err.what()) + '\n'),
```
slavexx, line 246: The same (about converting C-string to std::string)

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
ivanboika: Yeah, it was test check, just to verify that problem is here, I'll fix that

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
ivanboika: Yeah, it is, because after that we add another one symbol(char). It could corrupt message

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
slavexx: Very interesting... OK

DataAccess/DataAccess.Postgre/PostgreRepositories.cpp
slavexx: OK

PR #591, opened 2022-08-02
Refactoring of PGRepositoryManager

### LINKED ISSUE #594: Creating interface of manager in DB module
## Parent Task: 
#589

## Description:
Find the interface of PostgreRepositoryManager and cut it out, put interface in the new file. We must use this interface with other repository instances like LiteSQL

## Acceptance Criteria:
- Whole structure isn't destroyed
- We can use interface in other repository instances

Also we can see the same structure in the container and adapter parts

ivanboika:
Changed location of 2 classes/structures on their own files


slavexx (approved):

Orange-hanter (approved):

DataAccess/Public/Include/DataAccess/FutureResult.hpp
```diff
@@ -0,0 +1,30 @@
...
+     * @details Uses like common wrapper for std::future.
+     */
+    template <typename TFromAny>
+    struct FutureResult
+    {
+    private:
+        RawFuture _future;
+
+    public:
+        explicit FutureResult(RawFuture&& future) : _future(std::move(future)) {}
```
KliKonNik, line 18: Maybe it's not so important but the parameter names could be like [this](https://github.com/f1nal3/Juniorgram/wiki/%D0%9A%D0%B0%D0%BA-%D0%BF%D0%B8%D1%81%D0%B0%D1%82%D1%8C-%D0%BA%D0%BE%D0%B4-%D0%BD%D0%B0-CXX#:~:text=13.%20%D0%9F%D0%B0%D1%80%D0%B0%D0%BC%D0%B5%D1%82%D1%80%D1%8B,%D0%B2%20%D1%82%D0%B5%D0%BB%D0%B5%20%D0%BA%D0%BE%D0%BD%D1%81%D1%82%D1%80%D1%83%D0%BA%D1%82%D0%BE%D1%80%D0%B0.)

DataAccess/Public/Include/DataAccess/RepositoryRequest.hpp
```diff
@@ -0,0 +1,97 @@
...
+     * @details It is also stores a priority of this request.
+     */
+    struct RepositoryRequest
+    {
+    private:
+        ePriority   _priority;
+        RequestTask _task;
+
+    public:
+        explicit RepositoryRequest(ePriority priority, RequestTask& task) : _priority(priority), _task(std::move(task)) {}
```
KliKonNik, line 43: Look above

KanstantsinTaukach (approved):

DataAccess/Public/Include/DataAccess/FutureResult.hpp
ivanboika: I guess at that rule we have mistake because at the same time we have to name class fields with underscore, but it doesn't happen in the rule

ivanboika: Also, rule says that we can name parameter with the same name as field only if this isn't used in the body of constructor, and parameter is used in the initialiazer list of constructor, not body
I think this rule should provide a way about how to avoid misunderstandings of naming(mostly when names are equal), but there everythink is okey

KliKonNik (approved):

memorisecodead (approved):

memorisecodead (approved):

PR #617, opened 2022-09-07
Adding message editing handler of answer

### LINKED ISSUE #618: Adding message edit hander of answer
## Parent Task: 
#609
## Description:
We should to adding handler for accepting answer of editing message.
 Without this the answer is will be lost.

memorisecodead:
After than we should does to refactoring of Client

### DIFF
--- a/Client/Client.Core/Client.Core.Static/Client.hpp
+++ b/Client/Client.Core/Client.Core.Static/Client.hpp
@@ -162,6 +162,8 @@ protected:
     virtual void onChannelSubscribingListAnswer(const std::vector<uint64_t> subscribingChannelList);
     /// Channel delete Answer handler
     virtual void onChannelDeleteAnswer(Utility::ChannelDeleteCode channelDeleteCode);
+    /// Channel edit Answer handler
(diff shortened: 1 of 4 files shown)


Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -162,6 +162,8 @@ class Client
     virtual void onChannelSubscribingListAnswer(const std::vector<uint64_t> subscribingChannelList);
     /// Channel delete Answer handler
     virtual void onChannelDeleteAnswer(Utility::ChannelDeleteCode channelDeleteCode);
+    /// Channel edit Answer handler
+    virtual void onEditMessageAnswer(Utility::EditingMessageCodes reactionState) const;
```
ivanboika, line 166: Why const? I mean, others are not, but this is. Especially this should be const or others should be const however it doesn't happen

memorisecodead: Agree with you. We should need to changes others methods to const.

ivanboika (approved):

Gorba41kD (approved):

PR #620, opened 2022-09-09
Common changes for Client && ConnectionManager

### LINKED ISSUE #619: Common changes for Client && ConnectionManager
## Description:
For our vision, we should fixing some code smells here.
At work with tests, we struggle with some problems with override functions and incorrect signature here.
In order to correctly complete the work on the tests, it is necessary to correct some points.

## Acceptance Criteria:
1. Fixing code smells.
2. Does try signature functions for override in future.
3. Small Refactoring for documentation.

memorisecodead:


ivanboika (approved):

slavexx (approved):
Good

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
+    /**
+    * @brief Connect to server with IP(host) and Port(port).
+    * @param const std::string_view& host - for identifing and accepting ip address,\
+    *        const uint16_t port - for accepting & identifing of port.
+    */
     bool connectToServer(const std::string_view& host, const uint16_t port);
-    /// Disconnect from server
+
+    /**
+    * @brief Disconnect from server.
```
Orange-hanter, line 38: "This method stop connection to remote host. Also, it reset all context and jobs that related with Client. This method call in destructor, so in general way it is not used."
That what I would like to see, instead of visual changes.

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
+* @class Client
+* @brief Class for handling connection
+*/
 class Client
 {
 public:
-    /// Disconnect from server
-    ~Client();
+    /**
+    * @brief Client virtual destructor.
```
Orange-hanter, line 24: Why this destructor is virtual, what responsibility it has?? How it implemented in current implementation?

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
-    /// Returns true if connection is established
+    /**
+    * @brief Returns true if connection is established.
+    */ 
     [[nodiscard]] bool isConnected() const;
 
-    /// Sends a message to server
+    /**
+    * @brief Sends a message to server.
+    * @param const Message& message - initialization accepting message.
```
Orange-hanter: _initialization accepting message_ - what does that mean??? Have you even asked yourself?

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
+    */ 
     void send(const Message& message) const;
 
-    /// Incoming message queue
+    /**
+    * @brief Incoming message queue.
+    * @return Utility::SafeQueue<Message> - incoming queue.
+    */
```
Orange-hanter, line 61: 1. Move implementation to cpp part
2. Documentation...

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
+    /**
+    * @brief Incoming message queue.
+    * @return Utility::SafeQueue<Message> - incoming queue.
+    */ 
     Utility::SafeQueue<Message>& incoming() { return _incomingMessagesQueue; }
 
-    /// Handler
+    /**
+    * @brief Loop handler.
+    *   Checking the queue for the elements in it.
```
Orange-hanter: Loop that handle incoming messages. Every type of message have specific function handler (see here...).
A list of all message types can be found here

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,172 +11,269 @@
...
+    */ 
     void loop();
 
-    /// Ping a server
+    /**
+    * @brief Ping a server.
```
Orange-hanter: Message handler that process "message::...." type. 
This handler actually do nothing, only log message travel time.

Client/Client.Qt/ConnectionManager.hpp
```diff
@@ -6,7 +6,10 @@
 #include <Models/Primitives.hpp>
 #include "ServerInfo.hpp"
 
-/// Login status
+/**
+* @class enum class LoginState.
+* @brief LoginState class for gives condition.
```
Orange-hanter: What condition, where to use them?

Orange-hanter (changes requested):
This commit more about styleguide than "common changes". If you work around documentation, so please at least make it better.

Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -46,7 +45,7 @@ bool Client::connectToServer(const std::string_view& host, const uint16_t port)
             onDisconnect();
         });
     }
-    catch (std::exception& exception)
+    catch (const std::exception& exception)
```
Orange-hanter, line 58: What are you afraid of making link to exception constant?

Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -165,25 +169,25 @@ void Client::storeReply(const std::string& message, uint64_t channelID, uint64_t
...
-    networkMessage.mBody = std::make_any<Models::ReplyInfo>(ri);
+    networkMessage.mBody = std::make_any<Models::ReplyInfo>(replyInfo);
     send(networkMessage);
 }
 
 void Client::userRegistration(const std::string& email, const std::string& login, const std::string& password) const
 {
-    // Generating password's hash which are based on login. It lets us to insert different users
-    // with the same passwords.
+    /// Generating password's hash which are based on login. It lets us to insert different users
```
Orange-hanter: Why this documentation here but not in header?

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,174 +11,283 @@
...
+    * @brief Disconnect from server.
+    *        This method stop connection to remote host. \
+    *        Also, it reset all context and jobs that related with Client. \
+    *        This method call in destructor, so in general way it is not used.
+    */ 
     void disconnectFromServer();
 
-    /// Returns true if connection is established
+    /**
+    * @brief �onnection check.
```
Orange-hanter: typo

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,174 +11,283 @@
...
+    *        Returns true if connection is established.
+    */ 
     [[nodiscard]] bool isConnected() const;
 
-    /// Sends a message to server
+    /**
+    * @brief Sends a message to server.
+    * @param const Message& message - initialized message \
```
Orange-hanter: What mean initialized?
Where I could find info about Message?

Client/Client.Core/Client.Core.Static/Client.hpp
Orange-hanter: I could read the signature of the function and understand that this function return incoming Message queue. 
But I steal didn't any idea about what this message queue, where it from and why I should use it.

Client/Client.Core/Client.Core.Static/Client.hpp
Orange-hanter: You didn't notice that this function handler of specific message type. Please, extend documentation with this information.

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,174 +11,283 @@
...
+    * @brief Sending the info of the created channel
+    * @param channel name std::string variable
+    */
+    void createChannel(const std::string_view& channelName) const;
+
+    /**
+    * @brief Send a message to server
+    * @param message Message
+    * @param channelID Channel ID
+    */
```
Orange-hanter: Should I comment every line?
Documentation is very poor, in future we could get a lot of problems with client implementation

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -11,174 +11,283 @@
...
-     * @brief Send an registration request to server
-     * @param email User E-Mail
-     * @param login User Login
-     * @param password Password hashed
-     */
+    * @brief Send an registration request to server
+    * @param email User E-Mail
+    * @param login User Login
+    * @param password Password hashed
+    */
```
Orange-hanter: Do we have some pattern for incoming parameters? Where in defined?

Client/Client.Qt/ConnectionManager.cpp
```diff
@@ -13,7 +13,7 @@ void ConnectionManager::init()
 {
     Settings::getInstance().beginGroup("ConnectionSettings");
     std::string host = Settings::getInstance().value("address").toString().toStdString();
-    auto port = Settings::getInstance().value("port").toInt();
+    int port = Settings::getInstance().value("port").toInt();
```
Orange-hanter, line 16: What a heck you're doing at Qt?

Orange-hanter (changes requested):

Client/Client.Core/Client.Core.Static/Client.hpp
memorisecodead: Here is the guideline: 

"If a class has any virtual function, it must have a virtual destructor, and that classes not intended to be base classes or not intended to be polymorphic should not declare virtual destructors."

Client/Client.Core/Client.Core.Static/Client.hpp
memorisecodead: This is old documentation that was written before my time...

Client/Client.Core/Client.Core.Static/Client.hpp
Orange-hanter: So why I read this here, and not at documentation?

Client/Client.Core/Client.Core.Static/Client.hpp
Orange-hanter: Yeah, but you right now not deal with any difficult tusk, so why not to try figure out what this mean...

Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -118,14 +122,14 @@ void Client::askForChannelSubscriptionList() const
...
+void Client::deleteChannel(const std::string_view& channelName) const
 {
     Network::Message networkMessage;
     networkMessage.mHeader.mMessageType = MessageType::ChannelDeleteRequest;
 
-    std::string ri;
-    ri                   = channelName;
-    networkMessage.mBody = std::make_any<std::string>(ri);
+    std::string replyInfo;
+    replyInfo            = channelName;
```
slavexx, line 141: We can write it in one line

slavexx (approved):

Client/Client.Core/Client.Core.Static/Client.cpp
```diff
@@ -320,78 +320,89 @@ void Client::loop()
...
                 uint64_t clientID = 0;
                 onServerMessage(clientID);
             }
             break;
 
             case MessageType::ChannelListRequest:
             {
-                auto channels = std::any_cast<std::vector<Models::ChannelInfo>>(message.mBody);
-                onChannelListRequest(channels);
+                auto channelsLists = std::any_cast<std::vector<Models::ChannelInfo>>(message.mBody);
```
ivanboika, line 340: Is it really lists of channels or list of channels?

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -3,19 +3,17 @@
...
 {
 /**
 * @class Client
 * @brief Class for handling connection.
-* @details Sends requests to the server. \
-*   This class contains methods for sending messages to the server \
-*   and header methods for processing responses from the server.
+* @details Client establishes a connection to the server, 
+*   Sends and processes requests of a certain type of messages to the database. \
+*   Contains virtual method headers for handling responses from the server.
```
Orange-hanter: The application core, which is a typical client in a client-server architecture.
The class manages the connection and provides a communication interface to the server side by providing a set of methods (API).
Each such function defines the data and order of actions that are needed to perform some functionality.
However, an option is provided to send messages manually.(@see send)

See the list of API functions and their descriptions in the class methods.

Client/Client.Core/Client.Core.Static/Client.hpp
```diff
@@ -64,160 +61,148 @@ class Client
 
     /**
     * @brief Loop that handle incoming messages.
-    * @details Every type of message have specific function handler, \
-    *   which in the process of processing the incoming message from the server goes \
-    *   in a loop for further processing of the response from the client.
+    * @details Handler function to process the server response message \
+    *   for the subsequent response from the client.
+    *   (You can see the message types in the Message header file).
```
Orange-hanter, line 68: The central method in which most of the work of the client kernel takes place. 
The function is an infinite loop that waits for messages to appear in the message queue.
Messages are divided into types and a different handler is called for each type. 
In the case of a non-registered type, an issue is logged and the work continues.

Orange-hanter (approved):

Gorba41kD (approved):

PR #622, opened 2022-10-04
Now Conan free

Orange-hanter:


f1nal3:
Removing specific version might break our tool chain in a random way, since distributions of default software on different platforms might be different. So I'd rather leave the version specifier, but it is debatable.

The other thing is why do we only remove the version specifier for Windows builds?

I do not have a strong opinion on the topic, but I do think there should be consistency between the platforms.

slavexx (approved):

KliKonNik (approved):

f1nal3 (approved):

PR #639, opened 2023-03-17
634: One of the possible implementations of the new Interface

### LINKED ISSUE #634: Change DB connection properties interface
## Description:
We have at ServerBuilder.hpp code that parse arguments for postgre
```cpp
// form a string with the configuration for connecting to the database
        std::string dbOptions = "";
        for (const auto& [configName, configValue] : _arguments)
...
DataAccess::IAdapter::getInstance<DataAccess::PostgreAdapter>(dbOptions));
```
This method is not so good, so it is better to transfer the parsing logic to DataAccess as an interface and implement it individually for each case.

## Acceptance Criteria:
Program behaviour are didn't change

KliKonNik:
I'll be glad to see your comments


Orange-hanter:
What I see does not quite fit the described requirements. The builder should not influence the construction of D.A.
Try to start by extending the constructor of the postgre module. After we see which interfaces can be extracted

PR #644, opened 2023-03-25
Investigate a bug in the GCC Linux build. 

### LINKED ISSUE #633: Investigate a bug in the GCC Linux build.
## Problem:
GCC Linux Build is behaving strangely when running tests. It was noticed that when running Utility.Test, an error was thrown. 
Attached are some screenshots: 
image
image
After the rebuild, this problem didn't occur. 
image

## Description:
Find out what the problem is, and tell me why it happens.

memorisecodead:
thinking about this

### DIFF
--- a/Utility/Utility.Test/UtilityTimeTest.cpp
+++ b/Utility/Utility.Test/UtilityTimeTest.cpp
@@ -9,30 +9,19 @@ using std::chrono::system_clock;
 TEST_CASE("consoleLogTimestamp test")
 {
     std::ostringstream outputTime;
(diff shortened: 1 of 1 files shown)


f1nal3 (approved):
lgtm if it gets rid of that issue ;)

memorisecodead:
> lgtm if it gets rid of that issue ;)

I hope this helps. 
I changed the structure of the test a bit: if before we created the tm type for the time format and compared two different local variables, now we have one local variable from the safe_localtime method. I also restarted the build 18 times on linux - nothing crashed.
image
image

ivanboika (approved):

PR #645, opened 2023-03-27
Back/cryptography

### LINKED ISSUE #625: Investigate Encryption/Compression
tldr
Sec is good, compressed bytes go brrrrrr

slavexx:


f1nal3 (approved):

ivanboika (approved):

f1nal3:
Ok, so there was a reason I put the links in the way I did: if you look at the workflows document you'll notice it uses the same notation as you do. And the picture is not displayed [here](https://f1nal3.github.io/Juniorgram/md__docs__workflow.html). 
This means we should use both GH notation and Doxygen notation or at least one and that one would be Doxygen.

PR #647, opened 2023-03-29
Adding ArgParser Tests

### LINKED ISSUE #642: ArgParser Coverage
We're close to getting an 80% coverage for Juniorgram. And while this is a relatively loose metric of code quality, seeing them green checkmarks would be great, eh?
image

memorisecodead:
Conclusions on reserch coverage of the remaining 30 percent: there are methods on the ArgParser.hpp side that take other data types, the example parse_args can take a vector of strings, when in our variant we specifically parse arguments via argc and argv. 
Consequently, covering the other 30 percent is not a possibility, based on the implementation. 
Maybe I haven't considered something, I'm waiting for your thoughts on this.

### DIFF
--- a/Server/Server.Test/ArgParserTests.cpp
+++ b/Server/Server.Test/ArgParserTests.cpp
@@ -5,34 +5,34 @@ TEST_CASE("Constructor of ArgParser [ArgParser][Success]")
 {
     SECTION("Default constructor")
     {
-        std::vector<const char*> args = {"./juniorgram"};
+        std::vector<const char*> args {"./juniorgram"};
         CHECK_NOTHROW(ArgParser(static_cast<int>(args.size()), args.data()));
     }
 
     SECTION("Constructor with default arguments")
     {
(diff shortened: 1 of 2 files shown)


Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -125,7 +125,7 @@ class ArgParser
                 err.what(), 
                 Base::Logger::LogLevel::ERR
             );
-            std::exit(1);
+            throw;
```
ivanboika: I guess that's a temporary stuff rn. Shall we move 117-128 lines (where we parse arguments) into separate function and after use kinda returning codes like enums. It may help us in the tests. I understand that 'throw' is just here to make tests work and that's PR is about tests but this 'throw' in catch segment scares me a lot

f1nal3 (approved):

ivanboika (approved):

PR #651, opened 2023-04-03
Improving ArgParser.hpp

### LINKED ISSUE #650: Improving ArgParser
## At the moment, 
our implementation of the argparser is not exactly intuitive and has a number of problems of its own: data type overflow when parsing data types larger than the type of the argument, as well as not working error handling on overflow. 

example of current realization with type overflow: 
image

## What needs to be done?
Solve the overflow problem, as well as improve intuitiveness when working with the argsparser.

memorisecodead:
Fixed an overflow problem, as well as an intuitiveness correction in the use of parser arguments.
The argparse library has been updated to version 2.9.

### DIFF
--- a/Server/Server.Core/Server.Core.Static/ArgParser.hpp
+++ b/Server/Server.Core/Server.Core.Static/ArgParser.hpp
@@ -46,11 +46,6 @@ public:
                 auto port = static_cast<uint32_t>(std::stoi(value));
                 if (port > std::numeric_limits<uint16_t>::max())
                 {
-                    Base::Logger::FileLogger::getInstance().log
-                    (
-                        "Bad port value", 
-                        Base::Logger::LogLevel::ERR
-                    );
                     return std::to_string(65001);
                 }
                 return std::to_string(port);
@@ -63,11 +58,6 @@ public:
                 auto port = static_cast<uint32_t>(std::stoi(value));
                 if (port > std::numeric_limits<uint16_t>::max())
                 {
-                    Base::Logger::FileLogger::getInstance().log
-                    (
-                        "Bad DataBase port value", 
-                        Base::Logger::LogLevel::ERR
-                    );
                     return std::to_string(5432);
                 }
                 return std::to_string(port);
--- a/Server/Server.Core/Server.Core.Static/ArgParser.hpp
+++ b/Server/Server.Core/Server.Core.Static/ArgParser.hpp
@@ -7,7 +7,7 @@
 * @class ArgParser.
 * @brief This class parses arguments with which application is started.
 * @details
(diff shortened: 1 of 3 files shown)


Gorba41kD (approved):

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -4,52 +4,48 @@
...
+        _argParser.set_assign_chars("=");
         _argParser.add_description("Server side of juniorgram messenger. Just have a fun!");
         _argParser.add_epilog("Some parts of the program may be configured by default.");
 
         _argParser.add_argument("--serverport")
             .default_value<std::string>("65001")
             .help("Set the port value for server needs")
             .action([](const std::string& value) -> std::string {
-                uint16_t port = static_cast<uint16_t>(std::stoi(value));
+                auto port = static_cast<uint32_t>(std::stoi(value));
```
ivanboika, line 46: I don't see a reason to cast up to 32. I guess we are using tcp protocol, so max port is 2^16

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -125,7 +115,7 @@ class ArgParser
                 err.what(), 
                 Base::Logger::LogLevel::ERR
             );
-            std::exit(1);
+            throw;
```
ivanboika, line 118: Screenshot_2023-04-05-19-03-38-286_com android chrome
This rethrow will cause stop of application
I've read that throwing in catch section is ok, but it must be handled properly 
Tell me if I'm wrong

KliKonNik (approved):
I was also confused with that port limit check

ivanboika:
Also, I want to clearify one single moment. I've seen a couple of default params in arg parser and as I understand it, we can run our server by default without typing any of params. Am I right?

memorisecodead:
> Also, I want to clearify one single moment. I've seen a couple of default params in arg parser and as I understand it, we can run our server by default without typing any of params. Am I right?

Sure

ivanboika (approved):

Server/Server.Core/Server.Core.Static/ArgParser.hpp
memorisecodead: Based on previous implementation, if the port exceeded a limit of type uint16_t, the default port is returned.
But because we have a limit of type 16_t, provincializing the port with a value exceeding type 16_t will cause an overflow problem and the port will be initialized with a random value of type 16_t.
The question arises, why do we need a limit check if it does not work?
Consequently, this check now.

PR #652, opened 2023-04-05
Added new logic for saving and using input parameters

### LINKED ISSUE #634: Change DB connection properties interface

KliKonNik:


Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -0,0 +1,26 @@
...
+class SettingsManager
+{
+public:
+    explicit SettingsManager(const Settings& settings)
+        : _settings(settings)
+    {}
+
+    std::string GetConnectionOptions() const
+    {
+        return std::string(
```
memorisecodead: Now, ArgParser doesn't parse arguments with '=' char. After adding improve of the ArgParser, it does work

memorisecodead (changes requested):

memorisecodead:
It's looks pretty. You must finish it, let's see the candy

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include "ArgParser.hpp"
```
Orange-hanter: ArgParser is external dependencies, so in case it will be changed on other module we got gap here. ArgParser should know about Settings, not vice versa.

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    std::string ServerPort() const  { return _serverPort; }
+    std::string HostAddress() const { return _hostAddress; }
+    std::string DBName() const      { return _dbName; }
+    std::string DBPort() const      { return _dbPort; }
+    std::string DBUser() const      { return _dbUser; }
+    std::string DBPassword() const  { return _dbPassword; }
+
+private:
+    std::string         _serverPort;
```
Orange-hanter: Why do not use map here?

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,46 @@
...
+    {
+        _serverPort     = parser.getPair("--serverport").second;
+        _hostAddress    = parser.getPair("--hostaddr").second;
+        _dbName         = parser.getPair("--dbname").second;
+        _dbPort         = parser.getPair("--port").second;
+        _dbUser         = parser.getPair("--user").second;
+        _dbPassword     = parser.getPair("--password").second;
+    }
+
+    void SetServerPort(std::string serverPort)      { _serverPort   = serverPort; }
```
Orange-hanter: So why do we need set-methods, are we planning to change parameters in future?

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "Settings.hpp"
+
+class SettingsManager
```
Orange-hanter, line 13: Quite confusing is the division of responsibility between settings and settings manager. Forget about the connection string to the database for now. I would take out the keys for the map fields in the settings class. And the data itself would be stored in the manager

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: It's a good idea. I'll do it.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: These setters were created to investigate problems with Server.Test. It's temporary.

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: I think the best way to realize this construction is composition, not aggregation.
So: Settings will be a private field of Manager and will be a structure with public fields only. But all interaction between them will be under the control of Manager. And exactly Manager will be that entity who will be able to receive parameters and return their combination.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Maybe you're right. But why do we need a separate entity to store parameters? We can use the Map as a field of Manager.
In this case, the manager will be responsible to receive parameters from anywhere, keep them and return their combination with various ways. It is Encapsulation in its pure view.

We can discuss this at the meeting on Wednesday.

KliKonNik:
What do you think of [this ](https://github.com/f1nal3/Juniorgram/pull/652/commits/59bd46f0b9997c32b9c58c5efdb3f95e0064db66)implementation?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -46,16 +47,13 @@ class ServerBuilder
...
-        }
-        else
-        {
-            server->initRepository(std::move(_repository));
         }
+       
+        server->initRepository(std::move(_repository));
 
-        uint16_t port = static_cast<uint16_t>(std::stoi(_settingsManager.GetServerPort()));
+        uint16_t port = static_cast<uint16_t>(stoi(_settingsManager.GetServerPort()));
```
Orange-hanter: Move back std::

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -1,52 +1,31 @@
 #pragma once
 
-#include "ArgParser.hpp"
-
-#include <PostgreRepositoryManager.hpp>
+namespace Server::Builder
+{
+struct ParamType
+{
+    constexpr static char* ServerPort  = "--serverport";
```
Orange-hanter: https://godbolt.org/z/6bbxYn59c
Need const

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -1,52 +1,31 @@
...
-    void SetRepoManager(DataAccess::IRepositoryManager* repoManager) { _repoManager = repoManager; }
-
-    std::string ServerPort() const                      { return _serverPort; }
-    std::string HostAddress() const                     { return _hostAddress; }
-    std::string DBName() const                          { return _dbName; }
-    std::string DBPort() const                          { return _dbPort; }
-    std::string DBUser() const                          { return _dbUser; }
-    std::string DBPassword() const                      { return _dbPassword; }
-    DataAccess::IRepositoryManager* RepoManager() const { return _repoManager; }
+    const std::string& GetValue(std::string key) const { return _settings.at(key); }
```
Orange-hanter: What if key doesn't exist?

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -1,52 +1,31 @@
...
-    explicit Settings(const ArgParser& parser)
+    inline Settings& SetValue(std::pair<std::string, std::string> value)
     {
-        _serverPort     = parser.getPair("--serverport").second;
-        _hostAddress    = parser.getPair("--hostaddr").second;
-        _dbName         = parser.getPair("--dbname").second;
-        _dbPort         = parser.getPair("--port").second;
-        _dbUser         = parser.getPair("--user").second;
-        _dbPassword     = parser.getPair("--password").second;
+        _settings.emplace(value);
```
Orange-hanter: There may be a problem with double writing.

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -1,47 +1,37 @@
...
-
-        return dbOptions;
+        return _settings.GetValue(ParamType::ServerPort);
     }
 
-    std::string GetServerPort() const { return _settings.find("--serverport")->second; }
-
 private:
-    std::map<std::string, std::string> _settings;
+    Settings _settings;
```
Orange-hanter: Pointer or unique_ptr better

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -1,47 +1,37 @@
...
-                    }
-                }
-                dbOptions += "=";
-                dbOptions += configValue;
-                dbOptions += " ";
-            }
-        }
-
-        return dbOptions;
+        return _settings.GetValue(ParamType::ServerPort);
```
Orange-hanter: why not move all the routine casting business here?

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -7,26 +7,28 @@
...
-    }
+    ~ServerBuilder() = default;
 
-    inline ServerBuilder& setValue(DataAccess::IRepositoryManager* repoManager)
+    /**
+    * @brief Sets up a specific RepositoryManager
+    * @details Allows to set a specific RepositoryManager
+    * @warning This method is to MockTest only! It may disappear in future.
+    */
+    inline ServerBuilder& SetMockRepo(DataAccess::IRepositoryManager* repoManager)
```
Orange-hanter: This function can be used not only in the mock case. Look at it like a framework developer.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -41,45 +43,23 @@ class ServerBuilder
...
-        auto port = static_cast<uint16_t>(std::stoi(_arguments["--serverport"]));
+        auto port = static_cast<uint16_t>(stoi(_settingsManager.GetServerPort()));
         server->initConnection(port);
 
         return server.release();
     }
 
-    std::map<std::string, std::string>  _arguments;
-    RepoManagerUPtr                     _repository;
+    SettingsManager     _settingsManager;
```
Orange-hanter: pointers please

Orange-hanter (changes requested):

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
memorisecodead: This method should initialize any repositories, not specifically linked to a mock repository

memorisecodead (changes requested):

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Thanks for your attention. this's my fault.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Done

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: In the last commit I have the next edition:
_settings.insert_or_assign(value.first, value.second);

Orange-hanter: @KliKonNik look at this like at potential copy-paste issue. I could forget change key so method `insert_or_assign` just erase my previous correct parameter.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
Orange-hanter: Please read the previous comment before writing your own.

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: Done

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: Done

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
KliKonNik: Done.
It has been renamed and its description has been changed.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
KliKonNik: Done

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
KliKonNik: Done.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: I decided to use the std::map::try_emplace.
Done

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -7,32 +7,34 @@
...
-    }
+    ~ServerBuilder() = default;
 
-    inline ServerBuilder& setValue(DataAccess::IRepositoryManager* repoManager)
+    /**
+    * @brief Sets up a specific RepositoryManager
+    * @details Allows to set a specific RepositoryManager
+    * @warning For now, this method is only used for MockTest! This may be changed in the future.
+    */
+    inline ServerBuilder& SetRepoManager(DataAccess::IRepositoryManager* repoManager)
```
Orange-hanter: Pay attention to the naming.

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,41 @@
...
+};
+
+class Settings
+{
+public:
+    Settings() = default;
+
+    inline Settings& SetValue(std::pair<std::string, std::string> value)
+    {
+        _settings.try_emplace(value.first, value.second);
```
Orange-hanter: extra copying

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,41 @@
...
+    constexpr static char const* DBUser      = "--user";
+    constexpr static char const* DBPassword  = "--password";
+};
+
+class Settings
+{
+public:
+    Settings() = default;
+
+    inline Settings& SetValue(std::pair<std::string, std::string> value)
```
Orange-hanter: The names of the functions do not correspond to the rules

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+        _settings.try_emplace(value.first, value.second);
+        return *this;
+    }
+
+    std::string GetValue(const std::string& key) const
+    {
+        if (_settings.find(key) != _settings.end())
```
Orange-hanter: if (auto it = _settings.find(key); it != _settings.end()) //since c++17 https://en.cppreference.com/w/cpp/language/if
    return *it;

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    explicit SettingsManager(std::unique_ptr<Settings> settings) : _settings(std::move(settings)) {}
+    
+    void Reset(std::unique_ptr<Settings> settings) { _settings.reset(settings.release()); }
+
+    /**
+    * @brief Generates connection string
+    * @details In this case, for now, connection string to Postgre DB only
+    */
+    std::string GetConnectionOptions() const
```
Orange-hanter: The names of the functions do not agree with the naming convention

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "Settings.hpp"
+
+namespace Server::Builder
+{
+class SettingsManager
+{
+public:
+    SettingsManager() = default;
+
+    explicit SettingsManager(std::unique_ptr<Settings> settings) : _settings(std::move(settings)) {}
```
Orange-hanter, line 18: Reason

Passing a smart pointer transfers or shares ownership and should only be used when ownership semantics are intended. A function that does not manipulate lifetime should take raw pointers or references instead.

Passing by smart pointer restricts the use of a function to callers that use smart pointers. A function that needs a widget should be able to accept any widget object, not just ones whose lifetimes are managed by a particular kind of smart pointer.

Passing a shared smart pointer (e.g., std::shared_ptr) implies a run-time cost.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics

Server/Server.Exec/MainServer.cpp
```diff
@@ -5,18 +5,12 @@
...
     ArgParser parser(argc, argv);
-    auto server = ServerBuilder()
-                                 .setValue(parser.getPair("--serverport"))
-                                 .setValue(parser.getPair("--dbname"))
-                                 .setValue(parser.getPair("--hostaddr"))
-                                 .setValue(parser.getPair("--port"))
-                                 .setValue(parser.getPair("--user"))
-                                 .setValue(parser.getPair("--password"))
-                                 .makeServer();
+    auto      server = ServerBuilder(std::make_unique<SettingsManager>(parser.GetSettings())).MakeServer();
```
Orange-hanter: There are nop sence for extra make_unique. GetSettings return smart pointer

Server/Server.Test/ArgParserTests.cpp
```diff
@@ -151,3 +155,27 @@ TEST_CASE("Method of obtaining a pair of arguments of the argparser [ArgParser][
         CHECK_THROWS_AS(parser.getPair("any_flag"), std::exception);
     }
 }
+
+TEST_CASE("Method for generating an object of the Settings class [ArgParser][Success]")
```
Orange-hanter, line 159: Please stop using tags worng!
https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#tags

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+class Settings
+{
+public:
+    Settings() = default;
+
+    inline Settings& SetValue(std::pair<std::string, std::string> value)
+    {
+        _settings.try_emplace(value.first, value.second);
+        return *this;
```
Orange-hanter, line 45: try_emplace does nothing if the item exists in the map. So here the code can rewrite the old entry. 
Write to the log, at least in the case of this scenario

Orange-hanter (changes requested):

Server/Server.Core/Server.Core.Static/Settings.hpp
memorisecodead: Why are you use keyword inline? 
It is best used with templates, or more specifically when specifying that it is a full specialty. 

Also if you want to make an exception to the ODR, but it's a class method, so I don't see the need.

Server/Server.Test/ArgParserTests.cpp
memorisecodead: Didn't we say that these tags are valid? If that has changed, let everyone know.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
KliKonNik: What do you mean exactly?
I think the method's name is good.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: corrected

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
KliKonNik: Oh, I got confused and I renamed all the methods according to the wrong rules.
I fixed it

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: @Orange-hanter I understood my fallacy and corrected it.

KliKonNik: @memorisecodead:
This keyword has been moved here from the old setValue method in ServerBuilder. At that moment, I thought it was strange for a class method. But I decided to leave it as is.
If someone else instead of me paid attention to this keyword, I delete it.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: It's a cool comment. I'll do so.
Thank you.

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: Done

Server/Server.Exec/MainServer.cpp
KliKonNik: Yeah, but parser.GetSettings() returns a std::unique_ptr<Server::Builder::Settings> for SettingsManager constructor and std::make_unique is colled for this SettingsManager.
It's correct.

Server/Server.Test/ArgParserTests.cpp
KliKonNik: I'll read "Test cases and sections" and "Tutorial" and then correct it.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: We have discussed the behavior of this method.
If a key is contained on map, a value won't be changed.
I'll add logging of  this situation. It's a good idea!

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: Hm... I'll think about it and try to use it.

Server/Server.Core/Server.Core.Static/Settings.hpp
memorisecodead: This would be better, because if necessary the compiler itself will make this method inline

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -124,12 +127,34 @@ class ArgParser
...
+    */
+    std::unique_ptr<Server::Builder::Settings> getSettings() const
+    {
+        auto getPair = [this](const std::string& str)
+                       {
+                           return std::pair<std::string, std::string>(str, _argParser.get(str));
+                       };
+
+        auto result = std::make_unique<Server::Builder::Settings>();
+        (*result).setValue(getPair(ParamType::ServerPort))
```
ivanboika, line 150: Why not ->?

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -0,0 +1,66 @@
...
+        {
+            return it->second;
+        }
+        return {};
+    }
+
+private:
+    std::map<std::string, std::string, std::less<>> _settings;
+};
+}  /// namespace Server::Builder
```
ivanboika: end of file

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -14,25 +15,49 @@ struct ParamType
...
+     *        In other words: pair<key, value>
+     */
+    Settings& setValue(const std::pair<std::string, std::string>& value)
     {
-        _settings.try_emplace(value.first, value.second);
+        if (!_settings.try_emplace(value.first, value.second).second)
+        {
+            Base::Logger::FileLogger::getInstance().log
+            (
+                "[Settings::setValue]: Write error! Parameter with key \"" + value.first + "\" already exists.",
```
Orange-hanter, line 41: In error case, next further work is doubt. You could switch to warning in otherwise

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -14,25 +15,49 @@ struct ParamType
     constexpr static char const* DBPassword  = "--password";
 };
 
+/**
+ * @class Settings
+ * @brief Class for collecting and storing parameters.
+ * @details Class can take a parameter, store it in a map and provide it.
```
Orange-hanter: Class container. Keep the required parameters for Server class configuration. Technically, it is a wrapper over the map.

Server/Server.Core/Server.Core.Static/Settings.hpp
```diff
@@ -14,25 +15,49 @@ struct ParamType
...
+ */
 class Settings
 {
 public:
     Settings() = default;
 
-    inline Settings& SetValue(std::pair<std::string, std::string> value)
+    /**
+     * @brief Setting up parameter method.
+     * @details Sets the value of <value.second> in the map with the key <value.first>.
```
Orange-hanter: Takes a pair as an argument and stores it in an inner container. In case of parameter duplication, skip new value and log about this situation

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
```diff
@@ -1,34 +1,54 @@
 #pragma once
 
 #include "Settings.hpp"
+#include "ArgParser.hpp"
 
 namespace Server::Builder
 {
+/**
+ * @class SettingsManager
+ * @brief Class for managing server builder settings.
```
Orange-hanter: Class provider of ServerBuilder settings. Contain high level logic that transform base parameters

Orange-hanter (approved):

Server/Server.Core/Server.Core.Static/ArgParser.hpp
KliKonNik: This is not  an attractive way of chaining commands with "->" in the first step and "." on other steps. And this syntax does not contradict generally accepted coding norms.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Done. Thanks

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: I used the next logic:
If you try to set a value with a used key, it wont a critical. Because the correct value is already set.
But your comment prompted me to look at this case from other angle.
It is strictly not recommended to try to set the value of the current Settings object by the used key.
Then the protocol error channel is more expected.

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Done

ivanboika (approved):

Server/Server.Core/Server.Core.Static/Settings.hpp
KliKonNik: Done.

Server/Server.Core/Server.Core.Static/SettingsManager.hpp
KliKonNik: Done.

memorisecodead (approved):

PR #654, opened 2023-04-09
Clean up server smells

memorisecodead:


Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -67,8 +67,8 @@ class ArgParser
             .default_value<std::string>("juniorgram")
             .help("set the name of database")
             .action([](const std::string& value) {
-                const std::vector<std::string> choices = {"juniorgram", "testdb"};
-                if (std::find(choices.begin(), choices.end(), value) != choices.end())
+                if (const std::vector<std::string> choices = {"juniorgram", "testdb"};
+                    std::find(choices.begin(), choices.end(), value) != choices.end())
```
KliKonNik, line 71: What is the reason for this decision?
Is the declaration and implementation in the conditional part a good practice?

Server/Server.Core/Server.Core.Static/ArgParser.hpp
```diff
@@ -79,8 +79,8 @@ class ArgParser
             .default_value<std::string>("127.0.0.1")
             .help("set the host address (optional argument)")
             .action([](const std::string& value) {
-                const std::vector<std::string> choices = {"127.0.0.1"};
-                if (std::find(choices.begin(), choices.end(), value) != choices.end())
+                if (const std::vector<std::string> choices = {"127.0.0.1"};
+                    std::find(choices.begin(), choices.end(), value) != choices.end())
```
KliKonNik, line 83: Same as above.

Server/Server.Core/Server.Core.Static/ArgParser.hpp
memorisecodead: C++17 introduced a construct to create and initialize a variable within the condition of if and switch statements and C++20 added this construct to range-based for loops. Using this new feature simplifies common code patterns and helps in giving variables the right scope.

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
```diff
@@ -47,7 +47,7 @@ class ServerBuilder
         {
             if (configName != "--serverport")
             {
-                for (size_t i = 0; i < configName.length(); ++i)
+                for (size_t i = 0; i < configName.length(); i++)
```
KliKonNik: The prefix increment is a better practice than the postfix.

P.S.: But in this case it's no matter. Because those lines will be removed.

Server/Server.Test/Utilities/TestUtility.hpp
```diff
@@ -115,11 +115,8 @@ inline Message& makeMessage(Message& message, MessageType messageType, MessageBo
                 message.mBody           = std::make_any<RegistrationInfo>(registrationInfo);
                 break;
             }
-            else
-            {
-                RegistrationInfo registrationInfo(mesgFiller.getRegistrationInfo());
-                message.mBody = std::make_any<RegistrationInfo>(registrationInfo);
-            }
+            RegistrationInfo registrationInfo(mesgFiller.getRegistrationInfo());
+            message.mBody = std::make_any<RegistrationInfo>(registrationInfo);
```
KliKonNik, line 119: It's the same logically.
But I would prefer the next realization:
        {
            if (messageBody == MessageBody::InvalidBody)
            {
                RegistrationInfo registrationInfo;
                registrationInfo._login = "";
                message.mBody           = std::make_any<RegistrationInfo>(registrationInfo);
             }
            else
            {
                RegistrationInfo registrationInfo(mesgFiller.getRegistrationInfo());
                message.mBody = std::make_any<RegistrationInfo>(registrationInfo);
            }
            break;
        }
Because thit way is more readable, understandable and unambiguous.
This is just a comment.

memorisecodead: I understand you perfectly, but there are too many if else blocks, as it reads much more complicated to me. There is a cognitive complexity from too many blocks:(

Server/Server.Core/Server.Core.Static/ServerBuilder.hpp
memorisecodead: to be honest in the course of the experiments I took out this method, accidentally corrected this way:)

ivanboika (approved):

f1nal3 (approved):

PR #688, opened 2023-05-20
667 clean up network smells

### LINKED ISSUE #667: Clean up Network smells
## Task:
It is necessary to "flush out" the network from the smells of the code.
I am attaching Sonor's analytics for smells analysis - [Sonar Cloud](https://sonarcloud.io/code?id=f1nal3_Juniorgram&selected=f1nal3_Juniorgram%3ANetwork).

## Criteria:
 Clean up the smells as much as possible, while maintaining the current functionality of the code.

Gorba41kD:


Network/Public/Include/Network/Handler.hpp
```diff
@@ -33,73 +35,42 @@ class Handler
     virtual ~Handler()            = default;
 
     /// Virtual Handler method for preprocessing next handler
-    virtual Handler*               setNext(Handler* handler)                                                      = 0;
+    virtual std::shared_ptr<Handler> setNext(std::unique_ptr<Handler> handler)                                                      = 0;
```
ivanboika: U have a unique as param and u return shared and I guess it's going to be the same Handler or not?

Network/Public/Include/Network/Handler.hpp
```diff
@@ -33,73 +35,42 @@ class Handler
...
-     * @brief Method for setting next handler fot message preprocessing.
-     * @param handler - next handler fot message preprocessing.
-     * @return handler - next handler fot message preprocessing.
-     */
-    Handler* setNext(Handler* handler) override
+    * @brief Method for setting next handler fot message preprocessing.
+    * @param handler - next handler fot message preprocessing.
+    * @return handler - next handler fot message preprocessing.
+    */
+    std::shared_ptr<Handler> setNext(std::unique_ptr<Handler> handler) override
```
ivanboika: Same as above

Network/Public/Include/Network/Handler.hpp
Gorba41kD: @ivanboika  If I return a unique pointer, the code will crash. It has to do with the architecture of this code. In the future, the protected object is used in method calls of this object. I did not change this implementation of the code, so I used a shared pointer. Because when returning a unique pointer, the protected object will be empty. And in the logic of this code, when calling methods, this protected pointer is used and it will be empty if return a unique pointer and the code will crash

ivanboika: Have you tried user a shared pointer as a param? Will it solve the issue?

memorisecodead: Incorrect method definition

Network/Public/Include/Network/SerializationHandler.hpp
```diff
@@ -389,7 +389,7 @@ class SerializationHandler : public AbstractHandler
     }
 
     template <typename T>
-    SerializedState processIncomingMessageBody(const yas::shared_buffer& bodyBuffer, Message& message)
+    SerializedState processIncomingMessageBody(const yas::shared_buffer& bodyBuffer, Message& message)const
```
memorisecodead: For beauty, make a space between the const and the body of the method

memorisecodead (dismissed):

Network/Public/Include/Network/Handler.hpp
Orange-hanter: remove this pseudo-improvement with link return. Keep the pointer in a unique reference. Make a generator function (you can create a separate file) that will combine all the handlers and spawn these handlers in itself

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -47,3 +47,4 @@ namespace Network
 		}
 	};
 }
+
```
Gorba41kD, line 50: The previous commit did not come to the remote repository, so I sent this test one, although it also did not immediately come. The git had problems...

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -0,0 +1,50 @@
...
+	{
+		public:
+		/**
+		* @brief Method for serialization, compression and encryption of outcoming messages.
+		* @param message    - buffer that contains data that should be serialized, compressed and encrypted.
+		* @param bodyBuffer - buffer that will contain serialized, compressed and encrypted body.
+		*/
+		MessageProcessingState operator()(const Message& message,yas::shared_buffer& bodyBuffer)
+		{
+			auto serializationHandler = std::make_unique<SerializationHandler>();
```
ivanboika, line 22: Does the old logic define this? I mean, is it so necessary to store those ptrs inside generator method? I just see that we're using in both methods same objects

Network/Public/Include/Network/Connection.hpp
```diff
@@ -73,22 +70,21 @@ class Connection : public std::enable_shared_from_this<Connection>
...
+    * message - "[connection id] Write Header Fail." is displayed.
+    */
     void writeHeader()
     {
         yas::shared_buffer bodyBuffer;
 
-        SerializationHandler handler;
-        handler.setNext(new CompressionHandler())->setNext(new EncryptionHandler());
-        MessageProcessingState result = handler.handleOutcomingMessage(_outcomingMessagesQueue.front(), bodyBuffer);
+        CombiningHandlers handler;
```
ivanboika: You're using twice the same object in different methods, I'd like to move 'em to private scope of class. Before they were two separate objects and it wasn't so important

Gorba41kD: @ivanboika , done.

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: @ivanboika , It's better to leave everything as it is. Because there may be a race condition. And also the following logical error occurs: an endless call to the next handler. It will be necessary to revise the architecture of the code.

ivanboika (approved):

Network/Public/Include/Network/Connection.hpp
```diff
@@ -68,6 +68,8 @@ class Connection : public std::enable_shared_from_this<Connection>
     Utility::SafeQueue<Message> _outcomingMessagesQueue;
     /* @details Buffer to store the part of incoming message while it is read. */
     Message _messageBuffer;
+    /* @details handler that is needed to combine all the handlers to preprocess the message. */
```
Orange-hanter: just handler that processes messages

Network/Public/Include/Network/Handler.hpp
```diff
@@ -4,18 +4,20 @@
 
 namespace Network
 {
-/** @enum MessageProcessingState
- *  @brief Successful or not result of message preprocessing
- */
+/** 
+* @enum MessageProcessingState
+* @brief Successful or not result of message preprocessing
```
Orange-hanter: flag showing the result of message processing

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include "Network/CompressionHandler.hpp"
+#include "Network/EncryptionHandler.hpp"
+#include "Network/SerializationHandler.hpp"
+namespace Network
+{
+	/** 
+	* @class CombiningHandlers
+	* @brief class that combines all handlers for message preprocessing.
+	*/
+	class CombiningHandlers
```
Orange-hanter, line 12: in such case make methods of this class static, so we could avoid useless memory allocation

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -0,0 +1,50 @@
...
+	{
+		public:
+		/**
+		* @brief Method for serialization, compression and encryption of outcoming messages.
+		* @param message    - buffer that contains data that should be serialized, compressed and encrypted.
+		* @param bodyBuffer - buffer that will contain serialized, compressed and encrypted body.
+		*/
+		MessageProcessingState operator()(const Message& message,yas::shared_buffer& bodyBuffer) const
```
Orange-hanter: Documentation is cool, but better way to talk to a developer is naming.
I guess the following signature looks much better:
`MessageProcessingState operator()(const Message& source, yas::shared_buffer& destination) const`

Orange-hanter (changes requested):

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: @Orange-hanter , done

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -14,36 +14,36 @@ namespace Network
...
+		* @param source      - buffer that contains data that should be serialized, compressed and encrypted.
+		* @param destination - buffer that will contain serialized, compressed and encrypted body.
 		*/
-		MessageProcessingState operator()(const Message& message,yas::shared_buffer& bodyBuffer) const
+		static MessageProcessingState handleOutcomingMessage(const Message& source,yas::shared_buffer& destination)
 		{
-			auto serializationHandler = std::make_unique<SerializationHandler>();
-			auto compressionHandler   = std::make_unique<CompressionHandler>();
-			auto encryptionHandler    = std::make_unique<EncryptionHandler>();
+			static auto serializationHandler = std::make_unique<SerializationHandler>();
```
Orange-hanter: The same handler for all messages?

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -14,36 +14,36 @@ namespace Network
...
 
 		/**
 		* @brief Method for decryption, decompression and deserialization of incoming message bodies.
-		* @param buffer - buffer that contains data that should be decrypted, decompressed and deserialized.
-		* @param message - variable that will contain decrypted, decompressed and deserialized message body.
+		* @param destination - buffer that contains data that should be decrypted, decompressed and deserialized.
+		* @param source      - variable that will contain decrypted, decompressed and deserialized message body.
 		*/
-		MessageProcessingState operator()(const yas::shared_buffer buffer, Message& message) const
+		static MessageProcessingState handleIncomingMessage(const yas::shared_buffer destination, Message& source)
```
Orange-hanter: Is buffer are destination here?

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -14,36 +14,36 @@ namespace Network
 		public:
 		/**
 		* @brief Method for serialization, compression and encryption of outcoming messages.
-		* @param message    - buffer that contains data that should be serialized, compressed and encrypted.
-		* @param bodyBuffer - buffer that will contain serialized, compressed and encrypted body.
+		* @param source      - buffer that contains data that should be serialized, compressed and encrypted.
+		* @param destination - buffer that will contain serialized, compressed and encrypted body.
 		*/
-		MessageProcessingState operator()(const Message& message,yas::shared_buffer& bodyBuffer) const
+		static MessageProcessingState handleOutcomingMessage(const Message& source,yas::shared_buffer& destination)
```
Orange-hanter: Name the function shortly, `make_handler`. Both functions could be named the same, but overloaded by arguments order

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: Yes, you are right. Better to stay buffer naming for this method

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: @Orange-hanter , Yes, and I already clarified this at the meeting and in a message to Ivan. If we want to move handlers to a private section, it is better to leave everything as it is. Here are the reasons: there may be a race condition. And also the following logical error occurs: an endless call to the next handler. It will be necessary to re-architect the code architecture. And you said that I could leave it as is.

Orange-hanter: I mean static qualifier. It make the same memory for all call every time. Is that what you trying to reach?

Network/Public/Include/Network/CombiningHandlers.hpp
Orange-hanter: You have yas buffer twice as destination point. Am I wrong, but at deserialisation process we handle buffer to message

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: Yes, to avoid reallocating memory on subsequent function calls.

Network/Public/Include/Network/CombiningHandlers.hpp
Gorba41kD: @Orange-hanter , Yes, you are absolutely right, the destination name for const yas::shared_buffer in the handlncomingMessage method does not make sense, so I suggested changing this name back to a buffer

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -31,19 +31,19 @@ namespace Network
 
 		/**
 		* @brief Method for decryption, decompression and deserialization of incoming message bodies.
-		* @param destination - buffer that contains data that should be decrypted, decompressed and deserialized.
-		* @param source      - variable that will contain decrypted, decompressed and deserialized message body.
+		* @param source      - buffer that contains data that should be decrypted, decompressed and deserialized.
+		* @param destination - variable that will contain decrypted, decompressed and deserialized message body.
 		*/
-		static MessageProcessingState handleIncomingMessage(const yas::shared_buffer destination, Message& source)
+		static MessageProcessingState makeHandler(const yas::shared_buffer source, Message& destination)
```
Orange-hanter: source send by value, investigate why

Network/Public/Include/Network/CombiningHandlers.hpp
```diff
@@ -0,0 +1,50 @@
+#pragma once
+#include"Handler.hpp"
```
KliKonNik: #include_[space]_"Handler.hpp"

KliKonNik (approved):

Orange-hanter (approved):

Network/Public/Include/Network/SerializationHandler.hpp
Orange-hanter: Problem resolved, do merge

PR #689, opened 2023-05-20
Network Tests

Gorba41kD:
Still need to write tests for Connection.hpp . And I don't know how to handle the case of throwing exceptions in YasSerializer.hpp. In methods serialize and deserialize.

### DIFF
--- a/Network/Network.Test/CompressionHandlerTest.cpp
+++ b/Network/Network.Test/CompressionHandlerTest.cpp
@@ -1,0 +1,23 @@
+#include "Network/Connection.hpp"
+#include "Network/Message.hpp"
+#include "Network/YasSerializer.hpp"
+TEST_CASE("CompressionHandler Test")
+{
+    yas::shared_buffer            bodyBuffer;
+    Network::Message              message;
+    SECTION("test method handleOutcomingMessage")
+    {
+        Network::CompressionHandler   compressionHandler;
+        REQUIRE(Network::MessageProcessingState::FAILURE == compressionHandler.handleOutcomingMessage(message,bodyBuffer));
+        compressionHandler.setNext(new Network::CompressionHandler());
+        REQUIRE(Network::MessageProcessingState::SUCCESS == compressionHandler.handleOutcomingMessage(message,bodyBuffer));
+        
+    }
+    SECTION("test method handleIncomingMessageBody")
+    {
+        Network::CompressionHandler   compressionHandler;
+        REQUIRE(Network::MessageProcessingState::FAILURE == compressionHandler.handleIncomingMessageBody(bodyBuffer,message));
+        compressionHandler.setNext(new Network::CompressionHandler());
+        REQUIRE(Network::MessageProcessingState::SUCCESS == compressionHandler.handleIncomingMessageBody(bodyBuffer,message));
+    }
+}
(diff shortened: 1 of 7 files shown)


Network/Network.Test/CompressionHandlerTest.cpp
```diff
@@ -0,0 +1,23 @@
...
+TEST_CASE("CompressionHandler Test")
+{
+    yas::shared_buffer            bodyBuffer;
+    Network::Message              message;
+    SECTION("test method handleOutcomingMessage")
+    {
+        Network::CompressionHandler   compressionHandler;
+        REQUIRE(Network::MessageProcessingState::FAILURE == compressionHandler.handleOutcomingMessage(message,bodyBuffer));
+        compressionHandler.setNext(new Network::CompressionHandler());
+        REQUIRE(Network::MessageProcessingState::SUCCESS == compressionHandler.handleOutcomingMessage(message,bodyBuffer));
```
memorisecodead, line 13: I don't understand how the first require is different from the second

memorisecodead: In both cases, the message is empty

memorisecodead (changes requested):

Network/Network.Test/CompressionHandlerTest.cpp
ivanboika: I've checked how it works and the body of message doesn't make any sence, bc at this point this logic that you're talking about is at the different handler (serialization). There that's important to have a valid body, but not in compression handler
@Gorba41kD tell me if I'm wrong bc I'm not so good with network

memorisecodead: What is the point of serialising an empty object? As in this test case, there is no difference between the first and second body.
It just checks the assigned states of empty objects with headers.

If we're talking about serialisation, then the serialization test involves manipulating data. Conventionally, I'll give you an example: yml to txt conversion.
A correct case test would check if the serialisation is correct, not the header state.

ivanboika: We're not serialising there, that's compressing handler
image
Difference appears there, when in second case we set next handler, and if it exists, it's okay
But ofc, when logic will come to serialisation it's going to be failure bc as u mentioned, that's empty message
Maybe that's test exists to cover all conditions in both compression/encryption
Also, you can see that @Gorba41kD does all preparation with network message and fills it because he tests logic of serialization handler
I guess everything is okay, but those tests maybe seems weird

Gorba41kD: Yes, Ivan is absolutely right. There is not important if we have empty message or not in these test cases of compression handler class. Here the case of passing to the nextHandler is considered (Encryption, or Serialization; Yes I use the following statement compressionHandler.setNext(new Network::CompressionHandler()); It may seem strange, but for the logic of this test, this option is fine). For example: if the next handler is empty (== null), then we return FAILURE from the method. And if we installed the following handler i.e. allocated memory to the nextHandler from the heap, then we will return SUCCESS. To summarize, we just check the pointer nextHandler for null.

memorisecodead: What serialization are you talking about? =)

memorisecodead: You do realize that you are not serialization anything, right? 
Checking the state has nothing to do with checking if the serialization is correct

